PR #2, opened 2019-01-04
Pydmap Read 

mts299:
First pydarn functionality! 
DmapRead allows for users to read in dmap files and obtain records in a list of dictionaries. 
The original code of dmap was taken from https://github.com/SuperDARNCanada/backscatter 
and updated to python 3.6 and improved upon for speed and readability. 

README is updated for installation notes and example code. 

To test: 
```python
import pydarn
dmap = pydarn.DmapRead('20170114.map')
dmap_data = dmap.read_records()
```

This PR is to focus on the DmapRead functionality and the README. 

Testing suite setup, performance test suite, DEVELOPERS.md, and DmapWrite are for other PRs


ksterne:
I know I'm not listed as a reviewer, but I couldn't help myself.  These are mostly nit-picky comments, but hopefully helpful.

Do we need to add numpy to `README.md` as a prerequisite since it's been added with this PR?  Should "Example" be changed to "Examples"?  And then under the first example, the "DMAP files are a SuperDARN" doesn't seem like a complete thought.  Is there more to that line that got misplaced?

I'll see if I can find time tomorrow to give it a whirl.

mts299:
@ksterne No worries, more the marrier for reviewing. Plus others might be busy with the Borealis project, and I would like to push this PR ASAP so I can start building off of it.

Sorry, where do you see that numpy has been added to this PR? 
I added numpy into the setup.py just in case it was not installed. So it should be installed on setup if the user does not have it. 
I did the same for the python yaml library, but in some cases, you also need to install the development code thus it being a prerequisite. 

Yes, I must have written that sentence half awake or near the end of the day. I will re-look over it tomorrow.  

Sounds awesome! Please let me know what you think. Currently, I am working on the DEVELOPERS guide, some extra documentation and how to set up a nice test suite without having large files.

egthomas:
Will support for grid files be included at some point?  Also, it looks like the `xtd` fields for power and spectral width are missing from the `mapfile_types` in `dicts_to_file`.

mts299:
Sorry, there is two branches with the same name. Please make sure you are using the one that has the complete README file that runs with the following test code I put in the description. 

@egthomas you are using/looking at the dmapwrite code which is not the scope of the this PR. I left a note above dmapwrite that I need to clean it up and review it. That is legacy code from backscatter. Sorry for the confusion.

mts299:
However, I forgot to add grid files in the unit tests so thank you for pointing that out @egthomas

egthomas:
Ah my mistake, thanks for the clarification!

setup.py
```diff
@@ -28,7 +28,7 @@
     # used to import the logging config file into pydarn.
     data_files=[('pydarn',['pydarn/logging_config.yaml'])],
```
asreimer: Although not really a change in this PR, I couldn't get pydarn to import without changing this line to:

    data_files=[('pydarn',['logging_config.yaml'])],

Apparently, data_files isn't the best thing to be using? This guy even says using package_data is bad and that one should use a MANIFEST.in file instead: https://github.com/pypa/sampleproject/issues/30#issuecomment-143947944

README.md
```diff
@@ -0,0 +1,73 @@
...
+DMAP files are a SuperDARN 
+
+```python
+import pydarn
+dmap_file = "./20180410.C0.sas.fitacf"
+damp_reader = pydarn.DmapRead(dmap_file)
+dmap_data = dmap_reader.read_records() 
+
+# dmap_data[record number][paramter name].value
+print(dmap_data[0]['beam'].value)
```
asreimer: Needs to be changed to:

    print(dmap_data[0]['bmnum'].value)

zuul
Ghostbusters reference XD

README.md
```diff
@@ -0,0 +1,73 @@
...
+#### Turn on debugging 
+
+```python
+import pydarn
+import logging
+
+pydarn_logger = logging.getLogger('pydarn').SetLevel(logging.DEBUG)
+
+dmap_file = "./20180410.C0.sas.fitacf"
+damp_reader = pydarn.DmapRead(dmap_file)
```
asreimer: Typo: "damp_reader" should be "dmap_reader"

README.md
```diff
@@ -0,0 +1,73 @@
...
+print(dmap_data[0]['beam'].value) 
+```
+
+#### Turn on debugging 
+
+```python
+import pydarn
+import logging
+
+pydarn_logger = logging.getLogger('pydarn').SetLevel(logging.DEBUG)
```
asreimer: Typo: "SetLevel" should be "**s**etLevel".

Really, why doesn't logging module use underscores for methods...

pydarn/logging_config.yaml
```diff
@@ -29,9 +29,9 @@ handlers:
         formatter: pydarn_fmt
```
asreimer: Why is the time format "2018-11-05 15:39:49,678" instead of using a decimal (example: 2018-11-05 15:39:49.678)? It doesn't matter that much I guess, but standard US/UK english uses decimals.

asreimer:
Hey @mts299! This looks great! Most of my comments are minor things like typos. I was able to get the example code to run.

One major comment I didn't include in my review: I couldn't get the testing to work, except for `test/pydmap_tests/test_datastructures.py`

For both `test_dmapread.py` and `test_dmapwrite.py`:

      File "test_dmapwrite.py", line 17, in <module>
        import test_data
    ModuleNotFoundError: No module named 'test_data'

And finally: `python tests.py` yields:

    ImportError: Failed to import test module: tests
      File "tests.py", line 18, in suite
        mod = __import__(t, globals(), locals(), ['suite'])
    ModuleNotFoundError: No module named 'pydarn.tests'

README.md
```diff
@@ -0,0 +1,73 @@
...
+		  `$ source <environment name>/bin/activate`
+        * Navigate to where you cloned pydarn:  
+		  `$ python3 setup.py install`
+    2. Install in the system (root privileges required):  
+		   `$ sudo python3 setup.py install`
+
+### Example 
+
+#### Reading DMAP file 
+DMAP files are a SuperDARN
```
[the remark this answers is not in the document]
mts299: Previously suggest by @ksterne. 

> It would be great if there was a specification published somewhere, so maybe you and Keith could write something up and then it could be referenced here?

I am not sure what you are trying to get at here. What do you mean by "specification published"?  Write what up? 

I think, for now, I was just trying to give an example of what the current state of pydarn could do. Later on, there will be more documentation with better user case examples. However, I don't want to get too far ahead of myself and just keep it simple for now.

README.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Example 
+
+#### Reading DMAP file 
+DMAP files are a SuperDARN 
+
+```python
+import pydarn
+dmap_file = "./20180410.C0.sas.fitacf"
+damp_reader = pydarn.DmapRead(dmap_file)
```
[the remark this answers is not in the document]
mts299: fixed

setup.py
mts299: This is within the scope of the PR since it is a problem for someone to install. I will look at this option and a couple others I saw a while back.

README.md
mts299: fixed

README.md
mts299: Most python libraries are actually like that. If `setLevel` is it's own inherent class to logging, it follows the PEP 8 style. However, I think it is more of the downfall of python's beastly-ness. If logging was implemented fairly early on (earliest I found was python 2.6) and PEP 8 wasn't enforced until later like 2.7 then the developers of that library probably just stuck with the old syntax for consistency and to prevent other libraries from breaking from them changing method names.  ¯\_(ツ)_/¯
  

Fixed the typo

asreimer:
Hmmm, the edit to my comment didn't save. I noticed your description after I posted that comment about the testing not working and editted my comment accordingly. 2am might have had something to do with it.

README.md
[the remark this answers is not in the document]
asreimer: You and Keith are basically the only 2 people on the planet that know DMAP now. This comment is just a suggestion about something that could potentially be a good thing to do in the future.

I'm not a wizard at github reviews yet, so I didn't understand what I was doing in requiring things vs. leaving comments.

mts299:
@asreimer Okay, I mean `test_dmapread.py` and `test_datastructures.py` should work. However, there is that change you need to do with `test_dmapread.py` for data files. Which I will bring up sometime soon so I can figure out the test suite :)

pydarn/logging_config.yaml
mts299: Apparently, that comma is built in with the python's `asctime` function. I can look at trying to change it but if it is not a big deal then I won't stress about too much right now. 

Personally, I thought it was a typo on my part :p

README.md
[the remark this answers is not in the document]
mts299: I see, sorry, jumped the gun on my part. 

Naturally, I take comments in reviews as something you have to do or discuss on doing especailly when you "request changes to be made." Maybe say this is a suggestion, or perhaps something to consider later is blah, for better clarity?

asreimer:
I tested everything using the examples in the README.md and they work fine. Good to go.

That setup.cfg and MANIFEST.in are slick. Really nice work @mts299!

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+        array_shape = [self.read_data('i', 4) for i in range(0, array_dimension)]
+        array_shape.reverse()
+
+        # if shape list is empty
+        if len(array_shape) != array_dimension:
+            message = "Error: Array shape {shape} could not be read."\
+                    " cursor: {cursor}".format(shape=array_shape,
+                                               cursor=self.cursor)
+            raise pydmap_exceptions.DmapDataError(self.dmap_file, message)
+        # TODO Explain where does slist come from?
```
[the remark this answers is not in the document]
kkotyk: I should mention that when we scanned all of our files the only param that was empty was slist.

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+                message = "Error: Array {index}-dimension size {size}"\
+                        " exceeds record size: {rec_size}."\
+                        "".format(index=i,
+                                  size=array_shape[i],
+                                  rec_size=record_size)
+                raise pydmap_exceptions.DmapDataError(self.dmap_file, message)
+
+        # We could use np.prod(array_shape) but the for loop has a better
+        # time performance.
+        total_num_cells = 1
```
kkotyk: Maybe I wrote this, but what is meant by cells? maybe a more clear name would be better here.

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+        # parse strings or dmaps into arrays the way it can for other
+        # types because it doesn't
+        # know the sizes. They have to be manually read the slow way.
+        # Because chars
+        # are encoded as hex literals, they have to be read one at a
+        # time to make sense.
+
+        # what is DMAP format? Can we get an example that has a
+        # list of DMAPs?
+        # Also why is char also included?
```
kkotyk: I think again since they are hex literals, but I can't recall exactly.

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+        elif data_type_fmt == 's':
+            byte_counter = 0
+            while self.dmap_bytearr[self.cursor + byte_counter] != 0 or self.cursor + byte_counter >= self.dmap_end_bytes:
+                # add 1 byte to byte_counter because a string is a list of chars
+                byte_counter += 1  # 1 is the number of bytes of a char
+            self.bytes_check(self.cursor+byte_counter, 'cursor', self.dmap_end_bytes, 'Total number of Bytes')
+
+            # format byte counter to a string format for unpacking
+            # WARNING: not using format for performance purposes
+            char_count = '%ss'%byte_counter
```
kkotyk: change % to .format

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+        DmapDataError
+            if trying to read a DMAP data type
+
+        See Also
+        --------
+        Struct.unpack : converts the bytes array values from a C structure to
+        python objects
+
+        See DEVELOPER_README.md for more information on DMAP data types and
+        Struct.unpack.
```
kkotyk: I'm not sure I ever tested numpy casting vs struct.unpack here. Perhaps some performance benefits in using numpy.

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+        Return
+        ------
+        np.ndarray
+            an numpy N-D array of string read in from the byte array
+        """
+
+        data = []
+        # FIXME: still need to find a method for testing this... Or a file
+        #        that contains a string/DMAP array
+        for dim_size in shape:
```
kkotyk: I don't think this is working correctly. if shape is (3,3,3) for example your final array is going to have 9 elements, whereas your 3x3x3 array should have 27 elements. I think this is why I used recursion for this method. This is also going to leave your cursor in the wrong spot. You are also appending these values to a flat array without reshaping before the return.

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+            Mismatch on the data type and the size specified in the byte array.
+        """
+        array = np.frombuffer(self.dmap_buffer, data_type_fmt, total_number_cells, self.cursor)
+        self.cursor += total_number_cells * data_fmt_bytes
+
+        return array
+
+
+# TODO: Clean up
+class RawDmapWrite(object):
```
kkotyk: Ignoring write until next PR

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+                's': STRING,
+                'q': LONG,
+                'B': UCHAR,
+                'H': USHORT,
+                'I': UINT,
+                'Q': ULONG,
+            }.get(fmt, None)
+
+
+# TODO: consider removing this function and get the user to call RawDmapWrite
```
kkotyk: Kinda disagree. Maybe this could be refactored, but this method has all the types and fields for the common SD formats. In theory all fields should be cast to these types when writing, so the user can use numpy or python types in their analysis without having to worry if their output is going to be compatible with RST.

test/pydmap_tests/test_dmapread.py
```diff
@@ -0,0 +1,325 @@
...
+import numpy as np
+import logging
+import collections
+
+import pydarn
+
+pydarn_logger = logging.getLogger('pydarn')
+
+# Test files
+rawacf_file = "./testfiles/20170410.1801.00.sas.rawacf"
```
kkotyk: I think these files should be trimmed to only a few records. We dont need full 2hr files.

test/pydmap_tests/test_dmapread.py
```diff
@@ -0,0 +1,325 @@
...
+        self.assertIsInstance(dm_records, collections.deque)
+        self.assertIsInstance(dm_records[0], collections.OrderedDict)
+        self.assertIsInstance(dm_records[2]['IMF.flag'], pydarn.pydmap.datastructures.DmapScalar)
+        self.assertIsInstance(dm_records[3]['stid'], pydarn.DmapArray)
+        self.assertIsInstance(dm_records[8]['IMF.flag'].value, int)
+        self.assertIsInstance(dm_records[10]['stid'].value, np.ndarray)
+        self.assertEqual(dm_records[3]['stid'].dimension, 1)
+        self.assertEqual(dm_records[0]['stid'].shape[0], 14)  # this will be file dependent... future working test project.
+
+    # TODO: Again dependent on the file used :/
```
kkotyk: Force a file with some corrupt records for testing. I would just read in a file, scramble the buffer a bit and then rewrite the binary.

test/pydmap_tests/test_dmapread.py
```diff
@@ -0,0 +1,325 @@
...
+        """
+        Test read_records on a corrupt file
+
+        Expected bahaviour: raises pydmap expection
+        """
+        dmap = pydarn.DmapRead(corrupt_file2)
+        with self.assertRaises(pydarn.pydmap_exceptions.ZeroByteError):
+            dmap.read_records()
+
+    # FIXME: have not gotten this test to work with the streaming. I believe it is an utf-8 problem
```
kkotyk: Unfortunate. When I was testing, this was the single most valuable test.

test/pydmap_tests/test_dmapread.py
```diff
@@ -0,0 +1,325 @@
...
+    #        self.assertRaises(pydarn.DmapDataError,pydarn.parse_dmap_format_from_stream, corrupted_dmap)
+    #        del(records)
+    #        del(randomizer)
+    #        del(corrupted_dmap)
+
+
+if __name__ == '__main__':
+    """
+    Runs the above class in a unittest system.
+    Roughly takes 467 seconds.
```
kkotyk: I wonder if its possible to use mp with unittest.

test/pydmap_tests/test_dmapwrite.py
```diff
@@ -25,6 +25,9 @@
 import pydarn
```
kkotyk: Will review this for writing PR

README.md
```diff
@@ -0,0 +1,73 @@
...
+The following example shows how to read in a FITACF file, one of the SuperDARN's DMAP file types. 
+
+```python
+import pydarn
+dmap_file = "./20180410.C0.sas.fitacf"
+dmap_reader = pydarn.DmapRead(dmap_file)
+dmap_data = dmap_reader.read_records() 
+
+# dmap_data[record number][paramter name].value
+print(dmap_data[0]['bmnum'].value)
```
kkotyk: Is it possible to create a getter or some sort of decorator so that the user doesn't have to explicitly call .value on every data value they want? ex. if i call dmap_data[0]['bm_num'] then under the hood it returns me dmap_data[0]['bm_num'].value by default? I don't think this affects too much, but if you are writing a large script or something you are going to have a lot of extraneous looking code for .value.

egthomas:
I don't know if this is the appropriate place to post these links, but in one of the review or code comments I thought I saw the general question "What is dmap format?".  Here are a few short guides/tutorials/documentation written by the original developer (Rob Barnes) which may be of some general help:

https://superdarn.github.io/rst/superdarn/src.doc/rfc/0006.html
https://github.com/SuperDARN/rst/wiki/Using-DataMap-Files-In-IDL
https://superdarn.github.io/rst/general/src.lib/dmap/index.html

pydarn/exceptions/pydmap_exceptions.py
```diff
@@ -6,16 +6,81 @@
...
+                " at cursor = {cursor} does not exist in dmap data types."\
+                "filename: {filename}".format(name=data_name,
+                                              data_type=data_type,
+                                              filename=filename,
+                                              cursor=cursor)
+        super().__init__(self.message)
+        pydarn_logger.error(self.message)
+
+
+class ZeroByteError(Exception):
```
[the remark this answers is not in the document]
mts299: Personally, I like `Zero` because it is concise and something may have set it to zero which might not have made the bytes invalid. However, I get what you mean with the less than zero cases, so I added `NegativeByteError` which could indicate something really bad happened :/

pydarn/exceptions/pydmap_exceptions.py
```diff
@@ -6,16 +6,81 @@
...
+                " at cursor = {cursor}.".format(filename=filename,
+                                              element=element_info,
+                                              cursor=cursor)
+        super().__init__(self.message)
+        pydarn_logger.error(self.message)
+
+
+class MismatchByteError(Exception):
+    """
+    Raised if an element has <= 0 bytes.
```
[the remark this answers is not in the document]
mts299: Actually, the whole file had the wrong type of docstring style I decided to use. Woops! 

Fixed it now and re-wrote the `MismatchedByteError` description.

pydarn/pydmap/datastructures.py
```diff
@@ -0,0 +1,212 @@
...
+        scalar_list: list
+            A list of namedtuples following the DmapScalar structure
+        array_list: list
+            A list of namedtuples following the DmapArray structure
+        """
+        self.__record = OrderedDict()  # private to avoid setting the record.
+        self.num_scalars = 0  # may be useless attributes?
+        self.num_arrays = 0
+
+        # store scalars first into the
```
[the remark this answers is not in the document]
mts299: fixed

pydarn/pydmap/datastructures.py
```diff
@@ -0,0 +1,212 @@
...
+data_type : int
+    the DMAP numerical value for the data_type
+dimension : int
+    The number of dimensions of the array, e.i., 1D, 2D, 3D
+shape : list
+    A list of dimension sizes of the array
+"""
+class DmapArray(NamedTuple):
+    name: str
+    value: np.ndarray
```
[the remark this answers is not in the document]
mts299: First off I really would like to see a DMAP example where  DMAP records are stored in records. The Spec may be outdated or have legacy code that I would not like to introduce into newer code. Especially if the legacy introduces potential bugs. 

I am more than happy to go back if a given example is present in the current DMAP format.

With that said, `ndarrays` can hold any type of object including strings and DmapRecords so there is no actual issue here (The bittersweet of python :p ). It just informs developers that the object field should be a data type of ndarray. Technically if they store a list or something else nothing will really happen because it is python. Just later on in the downstream something might throw a `TypeError` but that is the pythonic way.  

I think in later python versions type hinting might actually enforce type checking. 

https://docs.python.org/3/library/typing.html

test/pydmap_tests/test_dmapread.py
mts299: If you read further into the file you see a FIXME which is indicating I could not get that rescambling stream unit test to work. Also it took 30 minutes to run  :/ 

Also, this PR is not focused on the test suite just yet because I need to figure out a way for the files to be reachable for anyone that wants to run the test suite. I did this merely to ensure that the code worked before pushing it out.

kkotyk:
@egthomas I'm not sure, but with Borealis, we usually use the issue tracker and use the description tags when we ask our questions or comments. "What is dmap format" would be a good issue with the question tag.

test/pydmap_tests/test_dmapread.py
mts299: Again not within PR scope, but you can use NOSE or the extension of unittests testtools to get concurrency. 

If you do it yourself you are asking for some inaccurate reporting because the print statements may not be thread-safe. 

Again out of scope and my test suite takes 6 seconds to run so currently not worth it for now. 

Will look into it when I focus on the testing Schema of this library.

test/pydmap_tests/test_dmapread.py
mts299: Why? 

I think the ability to read a DMAP file shows a lot of value and it to detect an error in bad files is just as valuable.

kkotyk: Because scrambling bits randomly hit almost all the test cases in where you could have corruption. I was able to use this method for hitting close to 100% code coverage.

kkotyk: This was important to me at the time since the realtime feed from prince george was crashing the backend every five minutes.

README.md
mts299: Personally, I like the `.value` because it is explicit and shows self-documenting code (best practices). It also keeps things consistent. Especially with the complaints that RST and DaVitpy have gotten with unclear variable names and code.  

For example: 
```python
for i in range(record[0]['beam'].shape[1] ):
      amplify *= record[0]['beam'].value
```
vs. 
```python
for i in range(record[0]['beam'].shape[1] ):
      amplify *= record[0]['beam']    
```

The former is clearer and easier to understand at first glance (which is me in debugging mode) and the latter is more confusing or black magic.  

Or to avoid extraneous looking code you can just store it in a variable ahead of time and then use it later to make the lines smaller. 

For Example: 
```python
Plot (record[0][velocity'].value,  record[0]['time'].value, default .... plot ... settings)
```
vs.
```python 
velocity_data = record[0][velocity'].value
time_data = record[0]['time'].value
Plot (velocity_data,  time_data, default .... plot ... settings)
```
The latter is also considered cleaner code too. I mean at the end of the day you are going to have to access the `record[0][velocity']` to get most of your data (assuming no loops are involved to cut some stuff down) so an extra `.value` especially with autocomplete seems okay to me :) 

Also, I don't know if you can make a getter function/decorator with a collection containing a collection. You are potentially looking at some overhead with a getter method and that means a decrease in performance.

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+        Reads an array from the byte array
+    read_string_DMAP_array(Shape, array_type)
+        Reads an array of strings/DMAP types into an array
+    read_data()
+        Reads the data type from the byte array
+    read_numerical_array(data_type, shape, total_number_cells)
+        Reads an array into a numpy array
+    """
+
+    def __init__(self, dmap_file: str, data_stream=False):
```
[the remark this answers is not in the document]
mts299: Not in 3.6 :) 

I use dmap_buffer later :/ ... I personally don't mind `file` since the main use case is to use a file.

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+            if len(dmap_file) == 0:
+                raise pydmap_exceptions.EmptyFileError("data stream")
+
+            self.dmap_bytearr = bytearray(self.dmap_file)
+            self.dmap_file = "stream"
+            pydarn_logger.debug("DMAP Read file: Stream")
+        self.dmap_buffer = memoryview(self.dmap_bytearr)
+        self.dmap_end_bytes = len(self.dmap_bytearr)
+        if self.dmap_end_bytes == 0:
+            raise pydmap_exceptions.EmptyFileError(self.dmap_file)
```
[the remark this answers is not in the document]
mts299: Except! if you look a bit higher in the code you can see I set self.dmap_file = stream log will just report stream and not the bytes.

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+
+    def zero_check(self, element: int, element_name: str):
```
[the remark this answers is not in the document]
mts299: done.

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+
+class DmapRead():
+    """
+    Reading and testing the integrity of DMAP files/stream.
+    ...
+
+    Attributes:
+    ----------
+    dmap_file : str
+        DMAP file/stream name/data
```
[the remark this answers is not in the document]
mts299: I will fix what I can but I will take a more serious look at this when I focus on DmapWrite again since I will go over the documentation again in io.py file to make sure things are updated.

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+            self.cursor += 4
+            block_size = self.read_data('i',4)
+            self.zero_check(block_size, "block size")
+
+            total_block_size += block_size
+            self.bytes_check(total_block_size, "total block size",
+                             self.dmap_end_bytes, "total bytes in file")
+
+            # why minus 2?
+            self.cursor = self.cursor + block_size - 2 * DMAP_DATA_TYPES[INT][1]
```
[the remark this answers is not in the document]
mts299: Later on, I realized that was the case but forgot to remove that comment. 

done.

pydarn/pydmap/datastructures.py
```diff
@@ -0,0 +1,212 @@
...
+            into the record.
+
+        See Also
+        --------
+        DmapScalar : named tuple for dmap array variable
+        """
+        if isinstance(scalar, list):
+            self._add_scalar_list(scalar)
+        else:
+            self.record[scalar.name] = scalar
```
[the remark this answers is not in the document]
mts299: Not exactly since _ means private but I know what you mean. 

Fixed!

pydarn/pydmap/datastructures.py
```diff
@@ -0,0 +1,212 @@
...
+        ---------
+        scalar: DmapArray/list
+            A DmapArray or a list of DmapArrays namedtuple that is added
+            into the record.
+
+        See Also
+        ---------
+        DmapArray : named tuple for dmap array variables
+        """
+        if isinstance(array, list):
```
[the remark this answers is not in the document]
mts299: Why not dict or tuples? 

In the docstring, I mention that it needs to be a list. Again bittersweet of python. Very non-pythonic to do type checking and to think all possible collections that I could get would just be extraneous code. I don't see the priority in this. 

Plus this class isn't actually used anymore, I just kept it just in case there was a reason to have it.

pydarn/pydmap/datastructures.py
```diff
@@ -0,0 +1,212 @@
...
+            self._add_array_list(array)
+        else:
```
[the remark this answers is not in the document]
mts299: fixed.

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+        pydarn_logger.debug("Reading DMAP records")
+        while self.cursor < self.dmap_end_bytes:
+            new_record = self.read_record()
+            self.__dmap_records.append(new_record)
+
+        self.bytes_check(self.cursor, "cursor",
+                         self.dmap_end_bytes, "total bytes in the file")
+        return self.__dmap_records
+
+    def read_record(self) -> collections.OrderedDict:
```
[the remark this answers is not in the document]
mts299: Actually, it depends, IO alone for parallelizing is a difficult challenge because typically you want to read things in order and write things in order and share a common file. Then you have to ask when parallelization is worth it? currently, on my computer alone it takes 1.945 s to read a 42MB RAWACF. If I add multiprocessing (mp) into the mix, you might not get any decrease in time because of the overhead to call mp and then create the processes (might actually increase the time). Also, you might need to know how many processes are available for the user's machine. Typically best practice is to optimize single process method first then look at concurrency if possible, and you have something to gain from it. 

However, I can see the benefit in parallelizing to read large full day FITACF files. Parallelization is possible to do since there are defined block sizes and with how modular the code is right now, it would be a couple of lines of code to change. One caveat is we probably would want the list to reflect the DMAP file so we would need to also spend some time on re-sorting the list. 

Good idea but more work needs to be looked into with how big does the file need to be to get a performance increase for DmapRead alone finally. Especially if @asreimer fixes cdmap then there is some exciting performance testing to be done :) 

In the testing branch, I am actually going to make a performance suite for testing performance of code :) (Request made by @asreimer)

Again out of scope and future want. So I will come back to this in the future. The main goal is to get the same capabilities as DaVitpy and then start adding in better performance methods and other features.

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+        #encoding_identifier = self.read_data('i',4)
+        self.cursor += 4
+        block_size = self.read_data('i',4)
+
+        pydarn_logger.debug("Reading Record {record}"
+                            .format(record=len(self.__dmap_records)))
+
+        # adding 8 bytes because code+size are part of the record.
+        # 4 is the number bytes for int format
+        remaining_bytes = self.dmap_end_bytes - self.cursor + 2 * 4
```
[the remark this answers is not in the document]
mts299: Good point especially since I did it earlier. 

Fixed.

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+        self.zero_check(num_arrays, "number of arrays")
+        self.bytes_check(num_scalars + num_arrays,
+                         "number of scalars + arrays",
+                         block_size, "block size")
+
+        pydarn_logger.debug("Reading record: reading scalars\n")
+        # originally called DmapRecord but then was just returning the ordered dict
+        # thus I am constructing it here instead for better speed performance.
+        record = collections.OrderedDict()
+        for i in range(num_scalars):
```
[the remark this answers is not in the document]
mts299: I don't remember, but I think it is just personal preference and I like `i` for better readability especially if people are not familiar with `_`

pydarn/pydmap/io.py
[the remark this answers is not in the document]
mts299: Not sure what point you are making, but do you know what slist is? 

@egthomas @ksterne any insight on if an empty array is valid or not?
Given the PR I was reviewing for @ksterne I was under the impression that if an array size = 0 we took it as invalid? 

We may need to raise this as an issue to have a further discussion.

pydarn/pydmap/io.py
mts299: You had `total_elements` and I felt elements from a mathematical standpoint was inaccurate. So I looked up matrix terms, and the number cells are equivalent to what you called the number of elements. Plus, I used elements earlier on as something else, so I didn't want break consistency.  

I left a comment to make it more clear.

pydarn/pydmap/io.py
mts299: Read the comment above the line. I use `%` because of a performance reason. 
`.format` is slower than `%` and since this method gets called on every parameter name the ms add up.

pydarn/pydmap/io.py
mts299: Legacy comments. 

But I forgot to check for DMAP types to raise an error, Fixed!

README.md
kkotyk: Sounds good. I think this is partly a preference thing and I don't think there are necessarily any right or wrong answers here. I see a couple of thumbs ups on my suggestion, so maybe some others can comment on what they think since this is ultimately a decision on top level functionality. Personally, I still stick by my suggestion as a preference provided there is no performance hits.

pydarn/pydmap/io.py
mts299: Nope, I tested them. 

Numpy only does well at reading in numerical arrays. If you get it to read in a single value or an array of strings similar to how you do it with struct.unpack it is slower and then on top of that numpy.frombuffer method returns it in an array so you need to call another method to get the string out or the single value which adds to the overhead :(

pydarn/pydmap/io.py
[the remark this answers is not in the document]
kkotyk: Maybe dmap_bytes? I know that file is the main use case and that does make sense, but it doesn't completely add up with the doc string and its not 100 clear to those who do work want to work with realtime streams. Maybe a clarification in the doc string then that file can represent any stream of bytes whether it be file or network?

pydarn/pydmap/io.py
mts299: Thus the FIXME comment. Once I have DmapWrite working, I can do some integration testing and try out an array of strings to figure out the loop structures. 

Currently, there is not a DMAP file type that has an array of strings, so I was unable to test this function correctly.

pydarn/pydmap/io.py
mts299: Out of scope. We shall see what happens when I clean up DmapWrite.

test/pydmap_tests/test_dmapread.py
mts299: Yes... again out of scope as this will be looked at in test suite branch.

pydarn/pydmap/io.py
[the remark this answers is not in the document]
kkotyk: Yeah I'm not 100% sure this is feasible. I do know that Pool() automatically determines number of available processors and that mp has a thing that will automatically preserve input ordering so that you don't need to rearrange, but I don't know about performance overheads.

pydarn/pydmap/io.py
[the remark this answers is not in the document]
kkotyk: slist is the array that holds the range gates that have valid data where qflg is 1. Its valid to be empty.

pydarn/pydmap/io.py
kkotyk: Sounds good. total_matrix_cells would work good as a name too.

test/pydmap_tests/test_dmapread.py
mts299: I can see that thus the FIXME tag. 

Will review when I get to focus on streaming with DmapRead. 

Realtime is out of scope for this PR and the test code that needs to be fixed.

pydarn/pydmap/io.py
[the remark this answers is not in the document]
mts299: `dmap_bytes` still makes no sense since it doesn't represent a file correctly. 

I didn't know I was making something for real-time streams thought this was a DaVitpy replacement which was not used in real-time streams :p 

I will update the documentation until a better suggestion comes along.

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+        if element <= 0:
+            element_info = "{name} {size}".format(name=element_name,
+                                                  size=element)
+            raise pydmap_exceptions.ZeroByteError(self.dmap_file, element_info,
+                                                  self.cursor)
+
+    def bytes_check(self, element: int, element_name: str, byte_check: int,
+                    byte_check_name: str):
+        """
+        Checks if the element greater than the byte_check.
```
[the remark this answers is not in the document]
mts299: It is not always used pertaining to the buffer. I re-worded it for better clarity. 

Fixed

pydarn/pydmap/datastructures.py
```diff
@@ -0,0 +1,212 @@
...
+        ---------
+        scalar: DmapScalar/list
+            A DmapScalar or a list of DmapScalars namedtuple that is added
+            into the record.
+
+        See Also
+        --------
+        DmapScalar : named tuple for dmap array variable
+        """
+        if isinstance(scalar, list):
```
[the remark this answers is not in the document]
mts299: See the above response on the same comment.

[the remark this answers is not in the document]
mts299: sorry below.

pydarn/pydmap/io.py
```diff
@@ -0,0 +1,1296 @@
...
+
+        if total_block_size != self.dmap_end_bytes:
+            message = "Error: Initial integrity check shows"\
+                    " total block size: {total_size} < end bytes {end_bytes}."\
+                    " Cursor: {cursor}."\
+                    " Data is likely corrupted"\
+                    "".format(total_size=total_block_size,
+                              end_bytes=self.dmap_end_bytes,
+                              cursor=self.cursor)
+            raise pydmap_exceptions.DmapDataError(self.date_file, message)
```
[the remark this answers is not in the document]
mts299: done

pydarn/pydmap/io.py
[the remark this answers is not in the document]
mts299: So any array can be empty and this okay?

kkotyk:
Other than maybe moving that data question to an issue or something, I think everything else is resolved. I can't submit an approval, so I'm just writing it here.

egthomas:
I don't know where it was in the 100+ comments, but somewhere I'd like to echo Keith's sentiment that ideally the dmap I/O routines should be as general as possible and therefore able to support real-time data streams and files with different formats from the standard SuperDARN rawacf, fitacf, etc.

mts299:
@egthomas I was not disagreeing, I have not tested streaming and focused more on file reading. I will be revisiting streaming in the test suite. There is also talk about supporting other file formats.

asreimer:
@egthomas we are all on the same page with this. And AFAIK, pydmap does still support data streams. @mts299 and I have been hashing out how to support multiple file types today.

Finally, can we please stop commenting on a closed PR and move this discussion into the chat?

egthomas:
@mts299 sorry I wasn't trying to imply anyone was disagreeing, just having trouble keeping up with all of the messages coming through my inbox.  This is my last comment on this PR request.

PR #5, opened 2019-03-06
Conversion functions

mts299:
Here is a smaller PR for review. Made some conversion functions to convert dmap records to dictionaries and vice versa. Might make the other PR easier to test as well. 

Fundamental functionality, further development will be to make SuperDARN specific conversion tools that cast the fields data to the correct types since python has some interesting caveats with data types. 

Also creating these functions to help with plotting functions.

### DIFF
--- a/pydarn/utils/conversions.py
+++ b/pydarn/utils/conversions.py
@@ -1,0 +1,124 @@
+# Copyright 2019 SuperDARN
+# Author: Marina Schmidt
+
+"""
+This module is to focus on data conversions
+
+All methods in this module are functions for
+quick conversions.
+"""
+from typing import List
+import numpy as np
+from collections import OrderedDict
+
+from pydarn import DmapArray, DmapScalar
+
+# key is the format char type defined by python, item is the DMAP int value for the type
+DMAP_FORMAT_TYPES = {'c': 1,  # character (char)
+                     'h': 2,  # short integer
+                     'i': 3,  # integer (int)
+                     'f': 4,  # float
+                     'd': 8,  # double
+                     's': 9,  # string
+                     'q': 10,  # Long int
+                     'B': 16,  # unsigned char
+                     'H': 17,  # unsigned short int
+                     'I': 18,  # unsigned int
+                     'Q': 19}  # unsigned long int
+
+# python possible types that could be the python format type characters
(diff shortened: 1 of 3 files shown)


pydarn/utils/conversions.py
```diff
@@ -0,0 +1,124 @@
...
+                     np.uint8: 'B',
+                     np.uint16: 'H',
+                     np.uint32: 'I',
+                     np.uint64: 'Q'}
+
+# !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+#       WARNING: This converts on best effort
+#       python does not have character types, if
+#       a string has a length of 1 then it will
+#       treated as a char.
```
asreimer: typo: **be** treated as char.

pydarn/utils/conversions.py
```diff
@@ -0,0 +1,124 @@
...
+            # if the value is an array then it is a DmapArray
+            if isinstance(value, list):
+                value = np.array(value)  # keep data structures consistent
+            if isinstance(value, np.ndarray):
+                shape = np.shape(value)
+                dimension = np.ndim(value)
+                # I did not implement the string length check to see if
+                # it is a list of chars because one would need to look at
+                # every element in the array, and it might be better to just
+                # store it as a string at this point.
```
asreimer: Nevermind the character encoding issues related to unicode vs. ascii, etc. Definitely something to revisit later, not right now.

asreimer:
I ran the unit test located at `test/test_utils/test_conversions.py` and it worked fine.

I also performed my own test as a sanity check with:

    import pydarn
    filename = '20171231.0001.00.sas.fitacf'
    data_handler = pydarn.DarnRead(filename)
    records = data_handler.read_fitacf() #reads all records
    records_indict = pydarn.dmap2dict(records) # test conversion to list of dictionaries
    records_fromdict = pydarn.dict2dmap(records_indict) # test conversion from dictionaries to named tuples

And this worked perfectly. I check that the records in each step and they converted as expected.

pydarn/utils/conversions.py
asreimer: Since it's so minor, I just fixed this myself so we can get this merged now.

asreimer:
I fixed the minor docstring typo myself. Testing is perfect. Merging.

PR #8, opened 2019-04-03
Pydarn's IO package

mts299:
Integration testing is complete, and performance scripts are written. Once again I haven't spent the time to make the trial smooth because of the need for DMAP files. 

Please take your time reviewing the code and testing it. 
Here are the performance times I got my computer:
Dmap module for 8.2 MB fitacf file:
Read: 4.027 s
Write: 1.716 s

SuperDARN io times: 
44 MB Rawacf file 
Read: 2.082 s
Write: 1.066 s

8.2 MB Fitacf file:
Read: 4.265 s
Write: 1.882 s

12 MB Iqdat file
Read: 0.096 s
Write: 0.55 s

12 MB Map file
Read: 1.653 s
Write: 0.735 s

1 MB Grid file
Read: 0.782 s
Write: 0.327

If you are wondering why there are changes in times based on file type oversize it is because of how many fields each file has and how many arrays. Reading arrays and strings are the biggest bottlenecks for performance.


pydarn/io/superdarn_formats.py
```diff
@@ -127,6 +125,10 @@ class Fitacf():
         'ptab': 'h',
         'ltab': 'h',
         'pwr0': 'f'}
+    extra_fields = {
```
aburrell: Would you mind adding comments above/at each of these variables that say what they are and specify the units they use?

mts299: Actually to save time and keep consistency with newly updated changes, I will add a reference (s) to where these variables are described in RST documentation.

kkotyk:
@mts299 Just wondering which part of this needs review. I notice there are a lot of changes and that some of this code has been in other PRs.

mts299:
Mainly the integration testing scripts. If the code changed it was just minor fixes for bugs I found.

test/integration/test_superdarn.py
```diff
@@ -0,0 +1,902 @@
...
+
+
+if __name__ == '__main__':
+    """
+    Runs the above class in a unittest system.
+    Roughly takes 467 seconds.
+    """
+    pydarn_logger.info("Starting DMAP testing")
+
+    unittest.main()
```
kkotyk: Should we maybe a test  in here for RST compatibility?

test/io_tests/test_pydmap_integration.py
```diff
@@ -0,0 +1,443 @@
...
+map_stream = "./testfiles/20170114.stream.map.bz2"
+iqdat_file = "testfiles/20160316.1945.01.rkn.iqdat"
+iqdat_stream = "testfiles/20160316.1945.01.rkn.stream.iqdat.bz2"
+grid_file = "./testfiles/20180220.C0.rkn.grid"
+grid_stream = "./testfiles/20180220.C0.rkn.stream.grid.bz2"
+
+pydarn_logger = logging.getLogger('pydarn')
+
+
+class IntegrationPydmap(unittest.TestCase):
```
kkotyk: No docstrings on methods.

test/integration/test_dmap.py
```diff
@@ -0,0 +1,302 @@
...
+import bz2
+import os
+
+import rawacf_data_sets
+import dmap_data_sets
+import copy
+from collections import OrderedDict
+
+# Test files
+rawacf_stream = "../testfiles/20170410.1801.00.sas.stream.rawacf.bz2"
```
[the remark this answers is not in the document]
kkotyk: *comment here applies to other files too

test/io_tests/test_pydmap_integration.py
```diff
@@ -0,0 +1,443 @@
...
+    def test_write_stream_read_fitacf(self):
+        fitacf_data = copy.deepcopy(fitacf_data_sets.fitacf_data)
+        fitacf_write = pydarn.DmapWrite()
+        fitacf_stream = fitacf_write.write_fitacf_stream(fitacf_data)
+
+        fitacf_read = pydarn.DmapRead(fitacf_stream, True)
+        fitacf_read_data = fitacf_read.read_records()
+        self.dmap_compare(fitacf_read_data, fitacf_data)
+
+#    def test_read_write_iqdat(self):
```
kkotyk: Can this commented out code be deleted?

test/performance/test_backscatter.py
```diff
@@ -0,0 +1,17 @@
+from backscatter import dmap
```
kkotyk: Is this a useful file to include for testing in the future?

test/performance/test_dmap.py
```diff
@@ -0,0 +1,44 @@
...
+import cProfile, pstats, io
+import logging
+import os
+
+"""
+This is a simple testing script for the Dmap read and write
+methods. Warning: need a file to read in.
+"""
+logging.getLogger('pydarn').setLevel(logging.INFO)
+#@profile
```
kkotyk: Instead of commenting this code out for mem profile, perhaps switch to a if __debug__: case or make some sort of conditionally dependent function.

kkotyk: Applies for further cases too.

test/performance/test_superdarn.py
```diff
@@ -0,0 +1,205 @@
...
+ps.print_stats()
+os.remove("test_dmap_performance.dmap")
+print("="*60)
+print("Write grid")
+print("File: ", grid_file)
+print("Size: {} MB".format(os.path.getsize(grid_file)/1000000))
+print("="*60)
+print(s.getvalue())
+
+dmap = pydarn.DmapRead(map_file)
```
kkotyk: These cases are pretty similar in structure. Can this duplicated code be worked into a function?

test/unit/test_dmap.py
```diff
@@ -0,0 +1,294 @@
...
+        self.assertIsInstance(dmap_records[0], collections.OrderedDict)
+        self.assertIsInstance(dmap_records[4]['bmnum'], pydarn.DmapScalar)
+        self.assertIsInstance(dmap_records[1]['ptab'], pydarn.DmapArray)
+        self.assertIsInstance(dmap_records[7]['channel'].value, int)
+        self.assertIsInstance(dmap_records[2]['ltab'].value, np.ndarray)
+        self.assertEqual(dmap_records[0]['ptab'].dimension, 1)
+        self.assertEqual(dmap_records[50]['gflg'].value[1], 0)
+
+    # TODO: Again dependent on the file used :/
+    def test_integrity_check_corrupt_file1(self):
```
kkotyk: I think it would be a good idea to just force a corruption of a file here. I think its possible to just read raw binary and then scramble a couple bytes

test/unit/test_superdarn.py
```diff
@@ -0,0 +1,1020 @@
...
+        Testing missing_field_check - Reverse idea of the extra_field_check,
+        should find missing fields in a record when compared to a key set of
+        SuperDARN field names
+
+        Expected behaviour
+        ------------------
+        Nothing - the missing_field_check should be able to handle subsets of
+        of complete sets. Meaning it will not raise an error if it contains
+        the full subsets of the dictionary keys but the full overall set.
+        This is needed for Fitacf, Gird and Map files as they do not write
```
kkotyk: > Gird

minor sp mistake

test/unit/test_dmap.py
mts299: I do that essentially in `def test_dmap_read_corrupt_stream(self):` similar to how you had it in your backscatter code. This was more can I detect an actual **blacklisted** file that is corrupt.

test/performance/test_backscatter.py
mts299: It was a comparison against yours for time performance, also a reminder of how to use your code. If you don't see a need for this test file then I can remove it.

test/performance/test_dmap.py
mts299: Personally, I prefer this way of logging. Just makes the code cleaner and leaves the logging library to figure things out. 

I personally also keep logging INFO on because this is something a user would be using as well.

mts299: Personally, I prefer this way of logging. Just makes the code cleaner and leaves the logging library to figure things out.

I personally also keep logging INFO on because this is something a user would be using as well.

test/performance/test_superdarn.py
mts299: Yes, I was thinking that since most tests are boilerplate code. However, I feel that should be a todo to refactor this code. I would like to this branch in `develop` soon rather than waiting for code refactoring.

mts299: added it to the todo's discussion :+1:

test/unit/test_dmap.py
mts299: I do that essentially in def test_dmap_read_corrupt_stream(self): similar to how you had it in your backscatter code. This was more can I detect an actual blacklisted file that is corrupt.

test/integration/test_superdarn.py
mts299: We could but that would definitely screw up `python test` since that would mean someone has to install RST. Maybe that is a good suggestion to ask reviewers to do.

test/integration/test_dmap.py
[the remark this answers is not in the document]
mts299: Maybe but I was also thinking of linking this to FRDR publication to keep files consistent otherwise testing values in the later tests or finding corrupt files may be challenging for a user to do thus my test would fail. I think unit tests should be self-contained so that you can actually predict and test behaviors for white box testing. 

Note: we don't include superdarn test files in this GitHub.

pydarn/io/superdarn_formats.py
```diff
@@ -3,6 +3,11 @@
 """
 This file contains several class pertaining to dmap format types
 that are used by SuperDARN.
+
+For description of these fields and their units see the following reference
+materials:
+    https://superdarn.github.io/rst/superdarn/src.doc/rfc/index.html
```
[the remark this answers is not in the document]
mts299: I think we already know RST documentation is severally outdated ... that is an RST revamp documentation problem. I don't think pyDARN should have to deal with that.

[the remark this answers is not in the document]
kkotyk: Okay, I just think this type of documentation should be somewhere since this IO stuff is extremely important. I don't think we should be too reliant on RST staying up to date. This can be done later.

test/integration/test_dmap.py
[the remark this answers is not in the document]
kkotyk: Alright, lets just add a TODO for clean up of this then. The current state does make it difficult for anyone to checkout the code and run the tests.

test/integration/test_superdarn.py
kkotyk: I think that will work.

test/performance/test_backscatter.py
kkotyk: Yeah I would say this can go. If its super critical later on we can always look back at the commits.

test/performance/test_dmap.py
kkotyk: Maybe Im just misinterpretting this, but I was commenting on the commented out `#@profile
#def mem_profile():` part. I was thinking this could be within a conditional instead of commented out. The logging is fine.

test/unit/test_dmap.py
kkotyk: Sounds good. I would suggest a more apt name might be something like "test_read_blacklist_file"

PR #14, opened 2019-06-17
Fixed missing elevation fields in FITACF files when Radars do not hav…

mts299:
Fixed the bug with `read_fitacf()` not being able to read in radar FITACF data that does not have elevation fields. 

This was a bug @aburrell found. 

Testing code:
```python
import pydarn

fitacf_file = ".20190203.0001.00.fhw.fitacf3"
fitacf_reader = pydarn.DarnRead(fitacf_file)
fitacf_dmap_data = fitacf_reader.read_fitacf()
```


aburrell:
I tested this using the example file, a different time from the same radar, and 20170228.2201.00.inv.fitacf (which has elevation angles).  It worked for all of the files.

mts299:
Angeline is pushing the button on my laptop :p

PR #17, opened 2019-07-17
Summary Plots

mts299:
This PR addresses a quick, simple line to plot a summary plot. 

Some things have changed since DaViT and DaViTpy, so I would like to get @ksterne and @aburrell perspective on usability and appearance of the summary plot. 

Things I changed: 
- [x] changed default color to be continuous
- [x] units in SI format 
- [x] color scheme default 'viridis' - colorblind-friendly 
- [x] presentation of the time-series plots 
- [ ] groundscatter only on the velocity plot 
- [ ] spacing between plots
- [ ] x-axis ticks on every plot for better mapping - Kathryn suggestion
- [ ] cp ID names - updated list 
- [ ] fitacf version number in the title 
- [ ] removal of fitacf generation command and time 
- [ ] adjustable figure size 
- [ ] removed saving to a file

Currently only supports range gates and no day-night terminator calculations. 

The plotting options can alter some of the above defaults: `color_map,` and title.

Examples:
Standard plot:
```python
import pydarn

fitacf_file = "20180220.C0.rkn.fitacf"
darn_read = pydarn.DarnRead(fitacf_file)
fitacf_data = darn_read.read_fitacf()

pydarn.RTP.plot_summary(fitacf_data, beam_num=0, figsize=(12,8))

plt.show()
```
normal_summary_plot

DaViTpy 
davitpy 20180220 rkn bm0

Borealis summary plot
```python
import pydarn

fitacf_file = "20190605.0000.02.sas.0.rawacf.dmap.fitacf"
darn_read = pydarn.DarnRead(fitacf_file)
fitacf_data = darn_read.read_fitacf()

pydarn.RTP.plot_summary(fitacf_data, beam_num=0, figsize=(12,8))

plt.show()
```
borealis_summary

No Elevation field
```python
import pydarn

fitacf_file = "20160301.C0.fhe.fitacf"
darn_read = pydarn.DarnRead(fitacf_file)
fitacf_data = darn_read.read_fitacf()

pydarn.RTP.plot_summary(fitacf_data, beam_num=0, figsize=(12,8))

plt.show()
```
summary_no_elevation

Let me know what you think and if you have any other suggestions. Otherwise, test and review as normal and merge :) 

Also, if we agree on the changes, I listed above I will add checkmarks to the boxes.


aburrell:
About to review and test, but one comment about colormaps.
1) I agree with you using the viridus colormap for all things that don't have a red-blue shift
2) Red/blue shift is an important physics aspect of our data and shouldn't be removed from our data visualisations.  It is also a pretty colorblind friendly color combo.  I'd recommend changing the velocity default colormap to 'RdBl' or another red/blue diverging colormap: https://matplotlib.org/users/plotting/colormaps/lightness_03.pdf

aburrell:
Another change, but less important:  I think it would be good to add overflow indicators to the colorbars whose ranges are restricted so that people are easily made aware of the saturation present in the plots.

aburrell:
Ok, I ran the test example and I get something different: the spectral width units are the old style instead of the new style shown in your figure.  However, the velocity units have been updated.

Also, despite the transmission frequency being present it isn't plotted :( . Or if it is, refer to my previous comment.

pydarn_summary_plot_test_inv

README.md
```diff
@@ -41,12 +41,27 @@ The following example shows how to read in a FITACF file, one of the SuperDARN's
...
+#convert to a dictionary
+fitacf_data = pydarn.dmap2dict(fitacf_dmap)
+
+#Do some parsing of the data!
+elv_list = []
+for fitacf_rec in fitacf_data:
+    if fitacf_rec['bmnum'] == 5:
+        elv_list.append(fitacf_rec['elv'])
+
+print(elv_list[100])
```
aburrell: Is this a debugging statement?  If so putting a comment here to that effect would be useful.

pydarn/plotting/rtp.py
```diff
@@ -1,6 +1,6 @@
 # Copyright (C) 2019 SuperDARN
 # Author: Marian Schmidt
-# This code is improvement based on rti.py in the DaVitpy library
+# This code is improvement based on rtp.py in the DaVitpy library
```
aburrell: This was correct before, it is based on rti.py davitpy, located at https://github.com/vtsuperdarn/davitpy/tree/master/davitpy/pydarn/plotting

pydarn/plotting/rtp.py
```diff
@@ -506,7 +489,226 @@ def plot_time_series(cls, dmap_data: List[dict], *args,
         ax.xaxis.set_major_formatter(dates.DateFormatter(date_fmt))
         ax.xaxis.set_minor_locator(dates.HourLocator())
         ax.margins(x=0)
-        return lines
+        ax.tick_params(axis='y', which='minor')
+        return lines, x, y
+
+    @classmethod
+    def plot_summary(cls, dmap_data: List[dict], *args, beam_num: int = 0,
```
aburrell: Question: why is `*args` placed where it is?

README.md
aburrell: Ok... was a bit lost in the reviewer window.  It looks now like this is example code and you're printing the output for obvious reasons.  If correct, just mark this as resolved :-P

mts299:
@aburrell I don't know what you are talking about obviously you need to pull more often to get those SI units :p 

Joking, sorry forgot to commit my code, try it now.

aburrell:
units now fixed!

pydarn/plotting/rtp.py
mts299: Funny store, a while back I was having issues calling a function with `args` and `kwargs` and thought it had to be positioned before the variable names, ends up not. Looks like it was one of those artifacts that happen when debugging but never went back to normal. 

I actually think we don't need args (I don't use the args) just the kwargs for the matplotlib plotting. So I will remove them :)

pydarn/plotting/rtp.py
mts299: Whoops replace and find got me there.

mts299:
After some of @aburrell comments here is what I had added:
- Changed the velocities color map to `RdBu`, however, because of the white background it doesn't look the best so I have added some other options if people want to mention their favorites
- added flexibility to change the color_map for any of the range_time plot 
- added overflow bars on the color bar 
- fixed the bug with `tfreq` 

New summary plot look:

default_color_map

To change various parameter color maps:
**magma**
```python
import pydarn

fitacf_file = "20180220.C0.rkn.fitacf"
darn_read = pydarn.DarnRead(fitacf_file)
fitacf_data = darn_read.read_fitacf()

# changing velocities color_map to magma 
pydarn.RTP.plot_summary(fitacf_data, beam_num=0, color_maps={'v': 'magma'}, figsize=(12,8))
plt.savefig('magma_color_map.png')
``` 
magma_color_map

**hot**
hot_color_map

**seismic**
seismic_color_map

**Spectral**
Spectral_color_map

**coolwarm** 
coolwarm_color_map


Color maps for matplotlib: 
https://matplotlib.org/users/colormaps.html

My thoughts/opinions: 
I like **hot** and **magma** most but it doesn't have the blue in it as scientists like to see. If we want blue and red my vote would be **Spectral** because you can kind of see the 0 velocity points. **coolwarm** and **RdBu** doesn't show the points as clearly and maybe meshed in with groundscatter if we keep that at *grey*.  

Thoughts on having `noise.sky` a red line instead of dotted for better visibility?

mts299:
@aburrell range on `tfreq` is fixed and fixed a bug with the legends in the line and tried out new line colors. 
Red it a bit better for when they do not overlap. Grey is still hard to see in comparison to black. 

tfreq_correction

Note: used Spectral for the velocity color map, soon realized not the best for color blindness. 

Also, realized none of the above color maps are good because they either have white in them (which is background color) or they are not the best for color blindness.

mts299:
Things I changed upon @aburrell comments:
- [ ] velocity color map `RdBu` - still undecided on the best one
- [x] `tfreq` default range set to 8-22 with minor ticks 
- [ ] changed line color for secondary scalar plots (`nave` and `noise.sky`) - still undecided on default
- [x] added overflow indicators on color bars

And fixed the miscellaneous bugs :p

aburrell:
Thoughts and options you can try based on your visibility/scientific clarity explorations:

- [ ] set the facecolor (background of the plot) to be '0.9' (very light grey) to improve the visibility of light colored points
- [ ] try `navy` and `skyblue`/`deepskyblue` for the line color in the primary/secondary scalar plots
- [ ] create a custom Red/Blue colorbar

I usually use `Spectral_r` with `set_facecolor('0.9')` but it isn't the best for colorblindness and I would like to use something better.  I used to use `seismic_r`, but it isn't great for intensity changes.  If we make a new one, and I think we should, I recommend we call it `doppler`.

Making your life even more difficult, it would be ideal to only include overflow on the parts of the colorbars where the data can overflow.  But I think this would make the code too cumbersome, so I am not putting it in the checklist, and was hesitant to bring it up.

mts299:
I like to keep consistent with things so if we set the background to 0.9 then I would do that for all the plots... which I am not sure how nice that would appear as. 

@aburrell you can also code up some examples to try different line colors too :) 

I think our best option is creating a new color scheme for pyDARN :) However, if that is the decision, then I vote that would be a new PR/issue, thus not hold up the preliminary summary plot work. 

In terms of overflow bars, I would either like to have them on all or no overflow at all. It keeps with consistency, and the user might get confused why some have them and not others (a.k.a they may think it is a bug). Sometimes it is hard just from the plot itself to know if there was overflow or not. 

Since velocity has overflowed, I would vote to keep them on. The overflow tabs don't take away from the information presented but show if there were overflow, then that would be what the color would extend into.

aburrell:
I agree that a new color scheme deserves it's own issue and pull request, so shouldn't hold this one up.

mts299:
I will add the warning even though this is matplotlib issue :p

mts299:
Added a warning message as suggested by @mardet987. 

@aburrell to not hold this PR too much, I have added a lot of flexibility in the code to change color schemes and line colors. Do you want to lift that "changes requested" on this PR? 

Currently going to default with `viridis` as it is well known and colorblind-friendly. I have added the option to change the color scheme for specific plots if people want to play around with it. 
For the colored lines, I don't think there is a better color choice in this case because no matter what the overlap kills it. However, you can also pick the line color, if you want to play around with it.

@asreimer and @ksterne do you have opinions/suggestions for the summary plots?

mts299:
@aburrell @asreimer @ksterne Does `nave` ever get higher than 40? 
@mardet987 and I are wondering if the scale should be 0-40 instead of 0-80.

egthomas:
> @aburrell @asreimer @ksterne Does `nave` ever get higher than 40?
> @mardet987 and I are wondering if the scale should be 0-40 instead of 0-80.

Yes, `nave` often does exceed 40 for 2-minute modes (ie 6+ second integration times).

ksterne:
Sorry I haven't been able to focus much energy/bandwidth here.  I'll echo what @egthomas notes here that yes it's possible `nave` is great than 40.  Though we don't run the 2-minute modes very often.  Not to influence it to look like our plots, but we use a 0-60 scale for our "summary" plots:
[http://vt.superdarn.org/tiki-index.php?page=Latest+Data](http://vt.superdarn.org/tiki-index.php?page=Latest+Data)

mts299:
Thank you @egthomas and @ksterne on your responses. I changed the default range for `nave` to 0-60. 

@ksterne I noticed on VT you have the frequency set to 8-18 instead of 8-22, do we transmit frequencies higher than 18 MHz? @aburrell you may know the answer to this as well. 

Also, @mardet987 noticed the `nave` was not being plotted correctly so I fixed that and here is the script I ran to show the fix: 
```python
import pydarn

fitacf_file = "20190605.0000.02.sas.0.rawacf.dmap.fitacf"
darn_read = pydarn.DarnRead(fitacf_file)
fitacf_data = darn_read.read_fitacf()

pydarn.RTP.plot_summary(fitacf_data, beam_num=0,
                        boundary={'nave': (10, 30)}, 
                        color_maps={'nave':'r'}, figsize=(12,8))
plt.show()
```

nave_correction

I want to change the title of plot a bit so hold off on merging.

aburrell:
So, the warning suggested by @mardet987 is fine the first time you plot something, but gets really annoying if you're making many of these or want to see what's going on in the script between plots.  I'd recommend piping this through Warnings or to stderr so that we don't see this Every Single Time we try and plot something.

mts299:
@aburrell Fixed the warning dpi issue, so you should be able to filter that dpi warning out now :) 

Also, I added a background option for summary and rang-time plots due to the overlap issue. Because of how color mesh lays pixels down, you need to color the `bad` (non-existent) points to the background color, which made `set_facecolor` useless with range-time plots. 

So now if you want to plot with a background-color this is what you need to do: 
```python
import pydarn

fitacf_file = "20190605.0000.02.sas.0.rawacf.dmap.fitacf"
darn_read = pydarn.DarnRead(fitacf_file)
fitacf_data = darn_read.read_fitacf()

pydarn.RTP.plot_summary(fitacf_data, beam_num=0, figsize=(12,8),
                        background_color='0.9')
plt.savefig('background_correction.png')
```

background_correction


With regards to the plotting color, nice job! However, because other people are focused on the new RST release and other things. Let's try and merge what we have now. Then before the actual pyDARN release we can have a teleconference on pyDARN and discuss the plotting standards include your color bar.

aburrell:
Great!  I'm good with that.  I tested the new behaviour, and it works as advertised.  Would you like me to merge?

mts299:
@aburrell yes, as long as you and @mardet987 are happy with this PR then we can merge it. Plus it is not going into `develop` so we still have time for bug fixes and other tests.

PR #19, opened 2019-08-26
Added additional data fields to superdarn_radars.py

DannoPeters:
geographic: Named Tuple containing geographic latitude longitude and boresight
geomagnetic: Named Tuple containing geomagnetic latitude longitude and boresight
decommissioned: date radar was decommissioned (shut down)
commissioned: date radar was officially commissioned (data considered stable)
beamSep: angular separation between radar beams


pydarn/plotting/superdarn_radars.py
```diff
@@ -1,101 +1,116 @@
...
-# Author: Marina Schmidt
+# Copyright (C) SuperDARN Canada, University of Saskatchewan
+# Authors: Marina Schmidt and Danno Peters
 
 """
 This module contains SuperDARN radar information
 """
 from typing import NamedTuple
 from enum import Enum
+from datetime import datetime
```
aburrell: Why not `import datetime as dt`?  Leads to a bit less confusion when reading the code.

pydarn/plotting/superdarn_radars.py
```diff
@@ -1,101 +1,116 @@
...
 
-
+"""
+_Hemispheres Class - Used to denote which hemisphere a radar is located in
+      _Hemisphere.North = 1
+      _Hemisphere.South = 0
```
aburrell: Ok, I'm going to suggest a change here and say that it should be North=1 and South=-1.  I feel very strongly about this, because it makes things easier when evaluating data; the sign of the hemisphere corresponds to the sign of the latitude.

pydarn/plotting/superdarn_radars.py
```diff
@@ -1,101 +1,116 @@
...
+"""
 class _Hemisphere(Enum):
     North = 0
     South = 1
```
aburrell: Doesn't match docstring.  Also, I think it should be changed.

pydarn/plotting/superdarn_radars.py
```diff
@@ -1,101 +1,116 @@
...
+_Radar Class - Used to store relevent information about the SuperDARN Radars
+      Nested inside of _Radar Named Tuple, composed of three floats representing location and boresight direction
+      
+      name: full text of radar name
+      stid: station number
+      abbrev: three letter station abreviation
+      beams: number of possible beams
+      gates: number of range gates per beam
+      geographic: Named Tuple containing geographic latitude longitude and boresight
+      geomagnetic: Named Tuple containing geomagnetic latitude longitude and boresight
```
aburrell: What about altitude?

pydarn/plotting/superdarn_radars.py
aburrell: Looking at other code, it seems that it is standard in this module to import datetime from datetime, so I am resolving this.

pydarn/plotting/superdarn_radars.py
mts299: Done.

pydarn/plotting/superdarn_radars.py
mts299: done

mts299:
Moving merge to `develop` since we don't have `plotting/rtp` anymore. I would like to also use this for my summary generation plots.

pydarn/plotting/superdarn_radars.py
mts299: Where do you get altitude values? 
@DannoPeters got this information from VT. 

Maybe we can add altitude in later?

mts299:
Will test this tomorrow, just cleaned up the file to meet PEP8 standards. 
If @aburrell if you are good with my suggestion on altitude then I will merge it tomorrow after my testing. 

I would like to move this to `develop` ASAP so I can use it in my summary plot generation scripts.

mts299:
Did this test in the python interpreter:
```
python 
Python 3.6.9 (default, Oct 29 2019, 10:39:36) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import pydarn 
>>> pydarn.SuperDARNRadars
<class 'pydarn.plotting.superdarn_radars.SuperDARNRadars'>
>>> pydarn.SuperDARNRadars.radars[65]
_Radar(name='Rankin Inlet', stid=65, abbrev='rkn', beams=16, gates=100, geographic=_Coord(lat=59.6, lon=-64.3, bore=-16.2), geomagnetic=_Coord(lat=72.6, lon=-26.4, bore=1.8), hemisphere=<_Hemisphere.North: 1>, institute='University of Saskatchewan', decommissioned=None, commissioned=datetime.datetime(2007, 5, 1, 0, 0), beamSep=3.24)
>>> pydarn.SuperDARNRadars.radars[65].beams
16
>>> pydarn.SuperDARNRadars.radars[65].geographic
_Coord(lat=59.6, lon=-64.3, bore=-16.2)
>>> pydarn.SuperDARNRadars.radars[65].hemisphere
<_Hemisphere.North: 1>
>>> pydarn.SuperDARNRadars.radars[12].hemisphere
<_Hemisphere.South: -1>
>>> pydarn.SuperDARNRadars.radars[12].name
'Syowa South'
```

mts299:
Merging code, and will be making an issue on adding the altitude to this file for the future.

pydarn/plotting/superdarn_radars.py
DannoPeters: I believe we need more accurate location data (VT's resolution is 2 decimal or 1.1km) before we can get accurate elevation data.

I think the highest resolution data set which covers all the radar sites would be the NASA Shuttle LiDAR (30m resolution and 16m accuracy). Here is a repo for parsing the elevation data set in python: https://github.com/tkrajina/srtm.py

egthomas: The VT radar table has been known to be incorrect many times in the past due to its reliance on manual updates.  The definitive source of radar latitude, longitude, and altitudes are the hdw.dat files, which provide altitude to an accuracy of a tenth of a meter (although you may only want to trust it to the nearest meter or so depending on the radar).  Furthermore, the geomagnetic latitudes/longitudes/boresites in the VT radar table are most certainly incorrect as those were found using the 2005 epoch of AACGM coefficients, which have been shown to be inaccurate by [Shepherd [2014]](https://agupubs.onlinelibrary.wiley.com/doi/full/10.1002/2014JA020264).

mts299: Created Issue #23 for this. Further discussion needs to be placed there.

PR #20, opened 2019-09-23
Range-time plotting module

mts299:
After the summary PR #17 I had to fix and refactor some of the code to make it more consistent. 
Passes all test_rtp.py unit tests. However, testing is still not setup smoothly but will be working on that in the near future so it is no file dependent or where you are running it. 

You can see my changes in the last 3 commits of the PR since it includes all commits from the three other PR's merged into this branch. 

The same tests as before but something new you can try is plotting RAWACF file! 
```python 
import pydarn
from matplotlib import colors

rawacf_file = "20190523.0200.00.sas.0.rawacf.dmap"
darn_read = pydarn.DarnRead(rawacf_file)
rawacf_data = darn_read.read_rawacf()
# AGB edited fit to fig
fig = plt.figure(figsize=(12, 8)) 
lognorm=colors.LogNorm
# AGB edited fitacf_data to rawacf_data
pydarn.RTP.plot_range_time(rawacf_data, beam_num=0, parameter='pwr0',
                           norm=lognorm, cmap='gnuplot')
plt.title("Rankin Inlet Fitacf 2.5 2018 Feb 20 00:01 - 21 00:00 Beam 0")
plt.xlabel('Date (UTC)')
plt.ylabel('Range gates')
plt.savefig('range-time_rawacf_pwr.png')
```
pwr_rawacf


kkotyk:
You're example lists that its plotting fitacf, but you have a rawacf file name as the arg. Also, I skimmed the PR #17 and the plotting scale is linear. Is log scale plotting intended here(I just haven't typically seen log scale plots with SD data)?

mts299:
Fixed the variable name typo, it is reading RAWACF if you read the code you can see the function `read_rawacf()` thus the reason I use clear function names :)

Your skimming must have been only for summary plots, range-time plots have this line: 
https://github.com/SuperDARN/pydarn/blob/698c556144e4a1c36b4a3266acc58bc0d83c2c55/pydarn/plotting/rtp.py#L303
Allowing the user to plot whatever normalization method they would like because flexibility in code is good practice to have and dependency injection is just beautiful :+1: 

Log scales are used in SD data, for power 0 in RAWACF file it is needed or else the plot will not look correct. We also use log scales for time-series plots for `noise`.
This example was also used to help analyze Borealis data :)

kkotyk:
Sounds good.

pydarn/exceptions/rtp_exceptions.py
```diff
@@ -0,0 +1,52 @@
...
+        pydarn_logger.error(self.message)
+
+
+class RTPUnknownParameter(Exception):
+    """
+    Error raised when the parameter is not found in the data passed in
+    """
+    def __init__(self, parameter):
+        self.parameter = parameter
+        self.message = "Error: The following parameter was not found in the data set. Please make sure it is typed correctly or you are using the correct data."
```
aburrell: Line should be broken up as in the previous classes so that it isn't so long.

pydarn/io/dmap.py
```diff
@@ -302,14 +311,14 @@ def test_initial_data_integrity(self):
         zero_check : raises ZeroByteError
         byte_check : raises MistmatchByteError
         """
-        pydarn_logger.debug("Testing the integrity of the dmap file/stream")
+        #pydarn_logger.debug("Testing the integrity of the /stream")
```
aburrell: Probably want to remove this or uncomment and write a more descriptive debug message.

pydarn/io/dmap.py
```diff
@@ -302,14 +311,14 @@ def test_initial_data_integrity(self):
...
         total_block_size = 0  # unit of bytes
         if self.cursor != 0:
             raise dmap_exceptions.CursorError(self.cursor, 0, self.rec_num)
 
         while self.cursor < self.dmap_end_bytes:
             """
-            DMAP files headers contain the following:
+            s headers contain the following:
```
aburrell: Typo

pydarn/io/dmap.py
aburrell: Here and elsewhere.

pydarn/io/dmap.py
```diff
@@ -893,32 +902,32 @@ def __str__(self):
     # HONEY BADGER method: Because dmap just don't care
```
aburrell: Cute, but more description needed.  Why doesn't DMap care?  Is it truly a honey badger?

pydarn/io/dmap.py
```diff
@@ -893,32 +902,32 @@ def __str__(self):
...
     def write_dmap(self, filename: str = ""):
         """
-        Writes dmap record to dmap file format.
+        Writes DMap record to  format.
 
         Parameters:
         -----------
         filename : str
-            The name of the DMAP file including path
+            The name of the  including path
```
aburrell: Typo (extra space)

pydarn/io/superdarn.py
```diff
@@ -35,10 +35,11 @@
...
 """
-from typing import Union, List
 import logging
 
-from pydarn import superdarn_exceptions, DmapRead, DmapWrite, superdarn_formats
+from typing import Union, List
+
+from pydarn import DmapRead, DmapWrite, superdarn_exceptions, superdarn_formats
 
 pydarn_log = logging.getLogger('pydarn')
```
aburrell: Ideally, `pydarn_log` or `pydarn_logger` would have the same name in all files.

pydarn/plotting/rtp.py
```diff
@@ -0,0 +1,942 @@
+# Copyright (C) 2019 SuperDARN
+# Author: Marian Schmidt
+# This code is improvement based on rti.py in the DaVitpy library
+# https://github.com/vtsuperdarn/davitpy/blob/master/davitpy
+
+"""
+Range-time Parameter plots (a.k.a Intensity)
```
aburrell: 'Time' should be capitalized and '(aka)' moved before 'plots'.

pydarn/plotting/rtp.py
```diff
@@ -0,0 +1,942 @@
...
+
+        # Default color maps for the summary plot
+        cmap = {'noise.search': 'k',
+                'noise.sky': 'k',
+                'tfreq': 'k',
+                'nave': 'k',
+                'p_l': 'viridis',
+                'v': 'viridis',
+                'w_l': 'viridis',
+                'elv': 'viridis'}
```
aburrell: This seems really inefficient.  How about adding another two keywords that would set the default line/color to black/viridis?

pydarn/plotting/rtp.py
```diff
@@ -0,0 +1,942 @@
...
+                    "".format(major=cls.dmap_data[0]['fitacf.revision.major'],
+                              minor=cls.dmap_data[0]['fitacf.revision.minor'])
+
+        radar_name = SuperDARNRadars.radars[cls.dmap_data[0]['stid']].name
+        # Date time formats:
+        #   %Y - year
+        #   %b - month abbreviation
+        #   %d - day
+        #   %H - Hour
+        #   %M - Month
```
aburrell: %M is 2-digit minutes of hour

pydarn/plotting/superdarn_cpid.py
```diff
@@ -0,0 +1,58 @@
+# Copyright (C) 2019 SuperDARN
+# Author: Marina Schmidt
+
+"""
+This module contains SuperDARN radar modes classified by cpid number
+Current documentation on modes:
+    http://superdarn.thayer.dartmouth.edu/info/modes.html
+"""
+
+#TODO: find/create better radar mode documentation
```
aburrell: Also frequency bands!

aburrell:
Well, I managed to cause this to crash horribly.

```
import pydarn
from matplotlib  import colors
rawacf_file = "20170214.0000.03.kod.c.rawacf"
darn_read = pydarn.DarnRead(rawacf_file)
rawacf_data = darn_read.read_rawacf()
fig = plt.figure(figsize=(12, 8))
lognorm=colors.LogNorm
pydarn.RTP.plot_range_time(rawacf_data, beam_num=3, parameter='pwr0',
                                                 norm=lognorm, cmap='gnuplot')
```
And you will get a very long crash flow that contains `ZeroDivisionError`

pydarn/exceptions/rtp_exceptions.py
mts299: Done!

pydarn/io/dmap.py
mts299: Woops for got to uncomment those in. 
Just to avoid repeated information, the debug information follows each other so you would get the file information in earlier messages. This one is indicating the function was called. Then later any error raised or other messages will be presented.

pydarn/io/dmap.py
mts299: Called an easter egg in code ;p Thus not in docstrings.

pydarn/io/dmap.py
mts299: done

pydarn/io/superdarn.py
mts299: changed

pydarn/plotting/rtp.py
mts299: done

pydarn/plotting/rtp.py
mts299: I don't understand your suggestion.

aburrell:
```
def routine(...., default_cmap = 'viridis', default_cline = 'k'):
```

and then use the defaults in the dict.

aburrell: That way you can change all the colors without having to remember all of the different variable names.

mts299: How about I check if cmap=str or a dict. if a str then all plots change and if a dict then you pick what changes.

mts299: because adding more options for the user to have to remember is also annoying and inefficient :p

aburrell: Whichever way you think is best.  From what I see, both methods have downsides.

mts299: Done

pydarn/io/dmap.py
mts299: fixed

mts299:
Fixed the bug, due to not being able to detect which type of normalization method is passed in I threw a more detailed error and how to solve it. Will be adding this into the tutorials for troubleshooting as well. 

The error you should get now is:
```bash
pydarn.exceptions.rtp_exceptions.RTPZeroError: Error: ZeroDivisionError or ValueError is raised due to trying to normalize 0.0 to 1273930.75 using the <matplotlib.colors.LogNorm object at 0x7fb10377d400> method for the parameter pwr0 at beam 3. This can be an issue with the data file, fields meaning something different. To resolve this error, please set the zmin or zmax parameter or try another normalization method.
```

This error will also appear for other `logNorm` exceptions/warnings when you try to plot FITACF `pwr0` data with `logNorm`

To fix the error in your case, you just need to set the `zmin` value: 
```
import pydarn
from matplotlib import colors

rawacf_file = "20170214.0000.03.kod.c.rawacf"
darn_read = pydarn.DarnRead(rawacf_file)
rawacf_data = darn_read.read_rawacf()
# AGB edited fit to fig
fig = plt.figure(figsize=(12, 8)) 
lognorm=colors.LogNorm
# AGB edited fitacf_data to rawacf_data
pydarn.RTP.plot_range_time(rawacf_data, beam_num=3, parameter='pwr0',
                           norm=lognorm, cmap='gnuplot', zmin=0.001)
plt.title("Rankin Inlet Fitacf 2.5 2018 Feb 20 00:01 - 21 00:00 Beam 0")
plt.xlabel('Date (UTC)')
plt.ylabel('Range gates')
#plt.show()
plt.savefig('ROS_DMap_pwr0.png')
```
ROS_DMap_pwr0

Also, here is an example of the new parameters for changing the summary plots color scheme
```python
import pydarn
from matplotlib import colors

fitacf_file = "20180220.C0.rkn.fitacf"
darn_read = pydarn.DarnRead(fitacf_file)
fitacf_data = darn_read.read_fitacf()
pydarn.RTP.plot_summary(fitacf_data, beam_num=0, cmaps='jet', lines='r')
plt.show()
plt.savefig('ROS_Ranking_summary_plot.png')
```
ROS_Ranking_summary_plot_example1

If you change the summary plot call to: 
```python 
pydarn.RTP.plot_summary(fitacf_data, beam_num=0, cmaps='jet', lines={'noise.sky': 'blue','noise.search':'red'})
```
ROS_Ranking_summary_plot_example2

aburrell:
This looks better.  I crashed the routine and got the improved error message.  When you write up the documentation, make sure to mention that you won't be able to re-use a figure handle once the routine crashes.  This is what you get:

ROS_DMap_pwr0_retry

After creating a brand new figure handle I obtained:

ROS_DMap_pwr0

Running the RTI plots using the two color options I get these plots.  This does speak to a need to have a larger (or adaptive) default range delta on the y-axis, but other than that are perfectly fine.

pydarn_summary_test_kod_1color
pydarn_summary_test_kod_2color

mts299:
@billetd Suggested rounding the timestamp on xmin to get the label to appear resulting in this:
ROS_Rankin_summary_plot_new

What it used to look like:
ROS_Rankin_summary_plot_old

Thoughts? Opinions?

aburrell:
As a default plot, I have no strong opinions.  I can judge the actual start time of the data equally well/poorly in both plots.

mts299:
What about this: 
ROS_Rankin_summary_plot_new


Side note: the change was only applied to the rang-time plots, whatever we agree on I will add it to the time-series plots.

billetd:
I was worried my opinion might be controversial! Though the argument was that people will typically read their average summary plot from left to right, so a less ambiguous start time could be beneficial (and also neater? Personal preference though...).

mts299:
@billetd No worries, others might be thinking it as well or prefer it. I think someone else like me the same question. This is a safe place to ask for personal preference and see what people think. Trust me the color maps are way more controversial. 

@aburrell fixed the range gates ticks:
hok_summary_plot

I will commit the code once we figure out the x-axis. @ecbland do you have any opinions on the matter?

aburrell:
Brill, thanks for changing the range gate range!

asreimer:
I prefer the RTP with the start time at the start of the x axis as per @billetd suggestion. These plots are looking fantastic by the way!

ecbland:
I also like @billetd's suggestion about the x-axis tick marks, but I don't feel super strongly about it :)
The plots are indeed looking very nice!

mts299:
Okay made the commits and allowed for the title to reflect the beams true start and end time to help the user read the plots. 

I also added some more instructions to the `RTPNoDataFound` exception:
```
pydarn.exceptions.rtp_exceptions.RTPNoDataFoundError: There is no Data for beam number 7 for the parameter type noise.search between the time range 2017 02 14 00:00 to 2017 02 14 02:00. Try another beam, for example, 3 or another parameter for the given time range.
```
So now if you read in a file that doesn't have the beam you are looking for it will suggest another possible beam you can try :+1:

pydarn/plotting/rtp.py
```diff
@@ -0,0 +1,988 @@
...
+        if isinstance(lines, str):
+            line.update({k: lines for k,v in line.items()})
+        else:
+            line.update(lines)
+
+        cmap = {'p_l': 'viridis',
+                'v': 'viridis',
+                'w_l': 'viridis',
+                'elv': 'viridis'}
+        if isinstance(cmaps, str):
```
aburrell: So....

This will fail if you have imported `unicode_literal`.  Probably better to test for the dict and let the mapping fail if the user puts in something horrible by mistake.

pydarn/plotting/rtp.py
```diff
@@ -0,0 +1,988 @@
...
+        title: str
+            title of the plot
+            Default: auto-generated by the files details
+            {radar name} Fitacf {version} {start_date} - {end_date}  Beam {num}
+        Raises
+        ------
+        IndexError
+        RTPUnknownParameterError
+        RTPIncorrectPlotMethodError
+        RTPNoDataFoundError
```
aburrell: This section seems unnecessary, but whatever.  ;-) . It does need a white line separating it from the input parameters.

pydarn/plotting/rtp.py
```diff
@@ -0,0 +1,988 @@
...
+        Plots the summary of the following SuperDARN parameter plots:
+            - noise.search : (time-series)
+            - noise.sky :  (time-series)
+            - tfreq : transmission frequency (time-series)
+            - nave : number of averages  (time-series)
+            - cp : control program ID (time-series)
+            - p_l : Signal to Noise ratio (range-time)
+            - v : velocity (range-time)
+            - w_l : spectral width (range-time)
+            - elv : elevation (optional) (range-time)
```
aburrell: This should all be in a notes section on the bottom (or in the returns description).

pydarn/plotting/rtp.py
```diff
@@ -0,0 +1,988 @@
...
+                    x.append(x[-1] + timedelta(0, 120))
+                    i = len(x) - 1  # offset since we start at 0 not 1
+                    if i > 0:
+                        z = np.insert(z, len(z), np.zeros(1, y_max) * np.nan,
+                                      axis=0)
+            # Get data for the provided beam number
+            if (beam_num == 'all' or dmap_record['bmnum'] == beam_num) and\
+               (channel == 'all' or
+                    dmap_record['channel'] == channel):
+                if start_time <= time:
```
aburrell: `time` is also a standard module, so should not be used as a variable name.

pydarn/plotting/rtp.py
mts299: Watch it, you might be showing some of your python 2 hairs : p Because python 2 and 3 handle strings different specifically UTF encoding. `unicode_literals` was developed to help python 2/3 backward compatibility. However, if a user was to use this and try to run my code in python 2 they will have a lot more problems. 

If we look at python3, however, 
you cannot directly import it:
```python
import unicode_literals 
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'unicode_literals'
```
If I use the `__future__` statement
```python 
>>> from __future__ import unicode_literals 
>>> name = 's'
>>> type(name)
<class 'str'>
>>> name = u's'
>>> type(name)
<class 'str'>
>>> name = u"s"
>>> type(name)
<class 'str'>
>>> name = "s".encode("utf-8")
>>> type(name)
<class 'bytes'>
```
So in this example the thing I would have to be concerned about would be bytes object and the user needs to deliberatly `encode` it to get that specific object. Since python isn't type set doing a million type checks is not very pythonic. So if does come up I can look at other possibilities. I say this because there is other `str` checks that cannot be changed like this one. 

For the sake of the feedback, however, I did reverse it as your suggestion, thank you for pointing out the potential issue with `bytes` :+1:

pydarn/plotting/rtp.py
mts299: Currently writing it with the numpy style guide, but I follow the numpy docstring style and prefer to include Raises because it allows non-code diving users to at least know what potential exceptions they may receive.

Fixed white space and added to the missing parameters in the above section.

pydarn/plotting/rtp.py
mts299: I think that is a personal preference. Personally I like it at the top as it details exactly what is being plotted as a descriptor should say and when users look at the docstring this would be the first thing they see. Where Notes would require some scrolling.

pydarn/plotting/rtp.py
mts299: Good catch changed to `rec_time`.

mts299:
I found my own error, got a RunTimeError with minor locator ticks because of the rounding down of the start time. With a 24 hr file, this exceeded the minute locator tick count. So I added a check to determine the interval on minute locator: 

rkn_summary_plot

cly_summary_plot

pydarn/plotting/rtp.py
aburrell: From the numpy style guide, the first line should be a: " one-line summary that does not use variable names or the function name"

mts299: Those are not variable names for summary plots and we don't need to be super strict with this. I think this classifies as high conflict vs. low reward. However, I don't want to hold up the merge any longer, so I will add this to the Notes as per request.

mts299: Done.

PR #42, opened 2020-04-22
Fix/borealis conversion error message

mardet987:
Just a quick fix PR - I'm working on an issue with files right now and found this error message misleading so updated it.


mts299:
Can you make this to develop, please?

mts299:
Looks good, not sure this needs to be tested since you just changed the wording around. 
@mardet987 what do you think?

mardet987:
Thanks for reviewing @mts299 ! I actually have done basic testing on this already since I was working on files that produced this error - it prints as expected! Let me know if you want anything else!

PR #45, opened 2020-05-22
RAWACF updates

mts299:
This PR focuses on fixing some bugs and updating SuperDARN Radars for getting new JME and DCN hardware files. 

**NOTE**: I am looking into a new method for getting radar info via radar.dat file so these things do not happen in the future. 

The RAWACF fixes where adding optional fields to read in various radar RAWACF files. 
One test that can be done is literally reading in DCN file as it some times is missing `acfd` field and that is apparently okay!

PR #46, opened 2020-05-22
FIX - sub repo bug

mts299:
Fixed the bug with the sub repo not getting the most recent master changes. Please review this PR before #45 as this may change those subrepo commits I did. Sorry about that!  

Test: 
Install pydarn from this branch
```bash
cd pydarn/utils/hdw
ls 
```
you should see the hardware file for DCN and JME :p


carleyjmartin:
Unless there are any objections, I'll merge so PR #45 can be tested.

PR #53, opened 2020-06-09
Time for MLT calculation

jpreistad:
Instead of using the time from the first beam in the file, use the time of the first beam in the scan specified upon call to fan.py (scan_index).


billetd:
Thank you for fixing this, jus sloppy work on my part! I’ll test today and merge.

PR #68, opened 2020-07-16
NEW Feature: Slant Range

mts299:
# Scope
Feature: `gate2slant`
module: *conversions.py*
package: *utils*

## Description
This new feature converts range gates to slant ranges (km) brought up by trying to overlay PFISR (Poker Flat Incoherent Scatter Radar) data with KOD SuperDARN data. 

I based the code off of [DaVitpy slant range calculation](https://github.com/vtsuperdarn/davitpy/blob/f2c11e34f781ba5fd027b5ce789d39ff1f4f86f1/davitpy/pydarn/radar/radFov.py)

## Test 
```python
import pydarn
import logging

#Load in dmap file
file = "20180220.C0.rkn.fitacf"
SDarn_read = pydarn.SDarnRead(file)
#Plot FOV and show
a = pydarn.RTP.plot_range_time(fitacf_data, beam_num=5)
#a = pydarn.RTP.plot_summary(fitacf_data, beam_num=5)
plt.show()
```
If you plot a Range-time Parameter plot you get:
slant1

If you plot a Summar plot you get:
slant2

## HELP!
- [ ] Validate the slant ranges are correct
- [ ] gate range 75 ~ 3500 km for summary plot purposes 
- [ ] Does 100 minor ticks make sense for slant ranges?
- [ ] should we keep 0 km as the axis point?

## TODO
- [ ] Documentation
- [ ] Unit Tests

### DIFF
--- a/pydarn/plotting/rtp.py
+++ b/pydarn/plotting/rtp.py
@@ -102,7 +102,7 @@ class RTP():
         ymax: int
             Sets the maximum y value
             Default: None, uses 'nrang' from data
-        slant: boolen
+        slant: boolean
             set the y-axis to slant range (km)
             if false will show gate numbers.
             Default: True
--- a/pydarn/plotting/rtp.py
+++ b/pydarn/plotting/rtp.py
@@ -642,7 +642,7 @@ class RTP():
     @classmethod
     def plot_summary(cls, dmap_data: List[dict], beam_num: int = 0,
                      groundscatter: bool = True, channel: int = 'all',
-                     slant=True, figsize: tuple = (11, 8.5),
+                     slant: bool = True, figsize: tuple = (11, 8.5),
                      watermark: bool = True, boundary: dict = {},
                      background_color: str = 'w', cmaps: dict = {},
                      lines: dict = {}, plot_elv: bool = True, title=None):
@@ -926,6 +926,10 @@ class RTP():
             else:
                 # Current standard is to only have groundscatter
                 # on the velocity plot. This may change in the future.
+                if slant:
+                    ymax = 3517.5
+                else:
+                    ymax = 75
                 if groundscatter and axes_parameters[i] == 'v':
                     grndflg = True
                 else:
@@ -942,7 +946,7 @@ class RTP():
                                         cmap=cmap[axes_parameters[i]],
                                         zmin=boundary_ranges[axes_parameters[i]][0],
                                         zmax=boundary_ranges[axes_parameters[i]][1],
-                                        ymax=3517.5,
+                                        ymax=ymax,
                                         background=background_color)
                 # Overwriting velocity ticks to get a better pleasing
                 # look on the colorbar
--- a/pydarn/plotting/rtp.py
+++ b/pydarn/plotting/rtp.py
@@ -55,8 +55,8 @@ class RTP():
                         zmin: int = None, zmax: int = None,
                         start_time: datetime = None, end_time: datetime = None,
                         colorbar: plt.colorbar = None, ymax: int = None,
-                        slant = True,
-                        colorbar_label: str = '', norm=colors.Normalize,
+                        slant: bool = True, colorbar_label: str = '',
+                        norm=colors.Normalize,
                         cmap: str = PyDARNColormaps.PYDARN_VELOCITY,
                         filter_settings: dict = {},
                         date_fmt: str = '%y/%m/%d\n %H:%M', **kwargs):
@@ -365,7 +365,11 @@ class RTP():
         if ymax is None:
                 ymax = max(y)
         ax.set_ylim(0, ymax)
-        ax.yaxis.set_ticks(np.arange(0, ymax+1, (ymax)/5))
+
+        if slant:
+            ax.yaxis.set_ticks(np.arange(0, ymax+1), 200)
+        else:
+            ax.yaxis.set_ticks(np.arange(0, ymax+1, (ymax)/5))
 
         # SuperDARN file typically are in 2hr or 24 hr files
         # to make the minute ticks sensible, the time length is detected
--- a/pydarn/plotting/rtp.py
+++ b/pydarn/plotting/rtp.py
@@ -14,7 +14,7 @@ from datetime import datetime, timedelta
 from matplotlib import dates, colors, cm, ticker
 from typing import List
 
-from pydarn import (dmap2dict, check_data_type, time2datetime,
+from pydarn import (dmap2dict, gate2slant, check_data_type, time2datetime,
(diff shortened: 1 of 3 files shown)


pydarn/utils/conversions.py
```diff
@@ -128,3 +128,46 @@ def dmap2dict(dmap_records: List[dict]) -> List[dict]:
...
+    if center:
+        # TODO: why -0.5? what does this value mean?
+        range_offset = -0.5 * record['rsep']
+    else:
+        range_offset = 0.0
+
+    # Now calculate slant range in km
+    slant_ranges = np.zeros(nrang+1)
+    for gate in range(nrang+1):
+        slant_ranges[gate] = (lag_first - record['rxrise'] +
```
mts299, line 171: Speed of light needs to be added somewhere?

billetd:
Looks good to me. My thoughts:

- I would have minor ticks every 200km as a default
- It would be more aesthetically pleasing if it rounded off the range on summary plots. Maybe to nearest 200?
- Happy with keeping y axis starting at 0

mts299:
@billetd Not sure I understand this point: 
> It would be more aesthetically pleasing if it rounded off the range on summary plots. Maybe to nearest 200?

But I set it to 200 for slant on summary plot so far: 
200_minor_ticks

If that looks good I will do the sam on plot_range_time

mts299:
nvm plot_range_time has it lol

billetd:
I like the rounding like that. That is what I originally meant, just so that there are no oddly specific numbers on the Y-axis.

mts299:
@carleyjmartin do you want to do a code review?

carleyjmartin:
> @carleyjmartin do you want to do a code review?

I'll have a look either later today or tomorrow

carleyjmartin:
Just testing with the data I have locally here. 

On import I got the old hardware file for dcn not found error.

I also noticed that when I want to turn off the slant range for the summary plots, so I set slant = False, it plots for range gates but the axis is in range gates:

code used 
```
import pydarn
import logging

#Load in dmap file
file = "/Users/carley/Documents/data/20131031.2201.00.sas.fitacf"
SDarn_read = pydarn.SDarnRead(file)
#Plot FOV and show
a = pydarn.RTP.plot_summary(fitacf_data, beam_num=5, slant=False)
#a = pydarn.RTP.plot_summary(fitacf_data, beam_num=5)
plt.show()
```

Screen Shot 2020-09-09 at 9 52 16 AM

This doesn't happen for range-time

pydarn/plotting/rtp.py
```diff
@@ -101,6 +102,10 @@ def plot_range_time(cls, dmap_data: List[dict], parameter: str = 'v',
         ymax: int
             Sets the maximum y value
             Default: None, uses 'nrang' from data
+        slant: boolen
```
carleyjmartin, line 105:
```suggestion
        slant: boolean
```

pydarn/plotting/rtp.py
```diff
@@ -920,10 +938,11 @@ def plot_summary(cls, dmap_data: List[dict], beam_num: int = 0,
                                         ax=axes[i],
                                         groundscatter=grndflg,
                                         channel=channel,
+                                        slant=slant,
                                         cmap=cmap[axes_parameters[i]],
                                         zmin=boundary_ranges[axes_parameters[i]][0],
                                         zmax=boundary_ranges[axes_parameters[i]][1],
-                                        ymax=75,
+                                        ymax=3517.5,
```
carleyjmartin, line 945: I guess this might be where you need an if statement, re: the comment from earlier?

pydarn/utils/conversions.py
```diff
@@ -128,3 +128,46 @@ def dmap2dict(dmap_records: List[dict]) -> List[dict]:
...
+            or edge
+
+    Returns
+    -------
+        slant_ranges : np.array
+            returns an array of slant ranges for the radar
+    """
+
+    # lag to the first range gate in microseconds
+    # TODO: with 2.0 and 0.3 what do these values mean?
```
carleyjmartin, line 154: I think the 2.0 refers to 'there and back' so the distance is doubled and the 0.3 is the speed of light in microseconds?

pydarn/utils/conversions.py
```diff
@@ -128,3 +128,46 @@ def dmap2dict(dmap_records: List[dict]) -> List[dict]:
...
+    lag_first = record['frang'] * 2.0 / 0.3
+
+    # sample separation in microseconds
+    sample_sep = record['rsep'] * 2.0 / 0.3
+
+    # Range offset
+    # If center is true, calculate at the center
+    if center:
+        # TODO: why -0.5? what does this value mean?
+        range_offset = -0.5 * record['rsep']
```
carleyjmartin, line 164: I think the -0.5 is so that it moves to the center of the range rather than the edge of it? I'm not sure though, it's the only logical thing I can think of.

pydarn/utils/conversions.py
carleyjmartin: See comment on review ;)

pydarn/utils/conversions.py
carleyjmartin: From units anyway lag(micro second) = frang (m) * 2.0 (unitless) / sol (m / microsecond).  

m's cancel leaving microsecond left.

carleyjmartin: Although I'm not sure 0.3 m/microsecond is the correct value? 
If the units are micro then c=300m/microsecond?
Is lag measured in nanosecond?

carleyjmartin: Never mind! It's in km right? That checks out!
lag(micro second) = frang (km) * 2.0 (unitless) / 0.3 (km / microsecond)

mts299:
@carleyjmartin try testing now

pydarn/plotting/rtp.py
mts299: yup fixed

pydarn/utils/conversions.py
mts299: so it looks good just will add 0.3 speed of light and 2 is for there and back. 

Should 0.3 be more accurate?

pydarn/utils/conversions.py
mts299: haha thanks!

pydarn/utils/conversions.py
carleyjmartin: I don't think you need it more accurate. It would be 0.2997..., which is going to give a very minimal difference in the actual slant range value you get imo.

carleyjmartin (approved):
All good from my end now!

mts299:
if you think it is good then we can merge it :)

PR #52, opened 2020-06-04
Fan plots

billetd:
# New Feature 
**Fan plots**

**module:** *fan.py*

**package:** *pydarn/plotting*

## Scope
This PR adds basic functionality for making fan plots from read in fitACF data. 

For a given scan number (dictated by the scan flags within the data list of dictionaries), either LOS velocity, power, spectral width or elevation can overlain on an AACGM latitude - magnetic local time plot. This code utilises the AACGMv2 library by @aburrell, making it now a requirement for pyDARN.

~~Currently, it does not allow for multiple different fans to be overlain on the same plot (either from the same radar or other)~~. It also does not currently support plotting the nearest scan to a given time (only scan numbers are used - i.e. the nth scan since the beginning of records). ~~Both of these features~~ this will be added in a future PR.

You plot your choice of range gates only by setting ranges=[two element list] of the lower and upper boundary. Default is ranges=[0,75].

Fan plotting also currently does not work for non standard beam-gate layouts (e.g. 15km range gates on Borealis radars) or special control programs (e.g. camping beams). It will always assume that the radar data given uses the standard number of beams and gates for that particular radar. For example, if you try to give it some Hankasalmi data with only 2 beams in the data strucutre (for whatever reason), it will plot something weird as it assumes it will be getting 16 beams for that radar. 

## Test code
```python
import matplotlib.pyplot as plt
import pydarn

#Read in fitACF file
file = "20200507.0.fitacf"
SDarn_read = pydarn.SuperDARNRead(file)
fitacf_data = SDarn_read.read_fitacf()

#Make the fan plot and show
fanplot = pydarn.fan.plot_fan(fitacf_data, scan_index=1, groundscatter=1, colorbar='True', parameter='p_l', colorbar_label='Power [dB]')
plt.show()
```

You might notice that the variable `fanplot` in the examples above actually holds some information. This contains the AACGM latitude and longitude of the fan just plotted, as well as the data and ground scatter information. If you instead change `fanplot` to 4 separate variables, it will return the latitude, longitude, data, groundscatter and datetime info into seperate numpy arrays:
```python
lats,lons,data,grndsct,datetime=pydarn.fan.plot_fan(fitacf_data, scan_index=27)

lats.shape

```
Which returns `>>>(76, 17)`, i.e. ranges x beams array of the AACGM latitude, and so on for the other variables. The groundscatter array is 0's and 1's, 1 being a range gate flagged as groundscatter.

[...]

PR #69, opened 2020-07-30
Scan locating and labelling

billetd:
# New Feature 
**Scan formation**

**module:** *scan.py*

**package:** *pydarn/utils*

## Scope
This PR adds rudimentary functionality to seperate a dmap_data list of libraries into scans, according to the scan flag specified within the records.

`build_scan` simply returns a 1D list with length equal to the number of dmap records, with an integer number indicating which scan each record is a part of. Initially this will be used for fan plots when merged, and so this functionality will be removed from PR #52. 

## Test code
```python
import pydarn

# Read in fitACF file. Works best with normalscan type data.
file = "20200507.0.fitacf"
SDarn_read = pydarn.SDarnRead(file)

# Sort into scans
beam_scan=pydarn.build_scan(fitacf_data)
print(beam_scan)
>>array([  1.,   1.,   1., ..., 720., 720., 720.])
```

### DIFF
--- a/pydarn/utils/scan.py
+++ b/pydarn/utils/scan.py
@@ -17,6 +17,11 @@ def build_scan(dmap_data: List[dict]):
     ----------
     dmap_data: List(dict)
         list of records (dictionaries) representing dmap data
+    Returns 
+    ----------
+    beam_scan: List
+        list of size equal to number of records in dmap_data, with scan number
+    for each record
     """ 
     # Set up scans for easy locating
     # Makes a list of size (number of records), with the scan number for each
--- a/pydarn/utils/scan.py
+++ b/pydarn/utils/scan.py
@@ -1,7 +1,7 @@
 # Copyright (C) 2020 SuperDARN Canada, Universtiy of Saskatchewan
 # Author(s): Daniel Billett
 """
-This module is is used for sorting a given dmap_data list of dictionaries
+This module is used for sorting a given dmap_data list of dictionaries
 into scans
 """
 
--- a/pydarn/utils/scan.py
+++ b/pydarn/utils/scan.py
@@ -26,13 +26,13 @@ def build_scan(dmap_data: List[dict]):
     # Set up scans for easy locating
     # Makes a list of size (number of records), with the scan number for each
     scan_mark = [sub['scan'] for sub in dmap_data]
-    no_scans = 0
+    current_scan = 0
     beam_scan = np.zeros((len(dmap_data)))
     for beam in range(len(dmap_data)):
-        if abs(scan_mark[beam]) == 1:
-            no_scans += 1
-            beam_scan[beam] = no_scans
+        if abs(scan_mark[beam]) == 1: #Absoloute value used due to some scan flags set as "-1"
+            current_scan += 1
+            beam_scan[beam] = current_scan
         if scan_mark[beam] == 0:
-            beam_scan[beam] = no_scans
+            beam_scan[beam] = current_scan
     
     return beam_scan
--- a/pydarn/utils/scan.py
+++ b/pydarn/utils/scan.py
@@ -1,0 +1,34 @@
+# Copyright (C) 2020 SuperDARN Canada, Universtiy of Saskatchewan
+# Author(s): Daniel Billett
+"""
+This module is is used for sorting a given dmap_data list of dictionaries
+into scans
+"""
+
+import numpy as np
+from typing import List
+
+def build_scan(dmap_data: List[dict]):
+    """
+    Returns list of size equal tonumber of records in dmap_data, with scan number
+    for each record
+
+    Parameters
+    ----------
+    dmap_data: List(dict)
+        list of records (dictionaries) representing dmap data
+    """ 
+    # Setup scans for easy locating
+    # Makes a list of size (number of records), with the scan number for each
+    scan_mark = [sub['scan'] for sub in dmap_data]
+    no_scans = 0
+    beam_scan = np.zeros((len(dmap_data)))
+    for beam in range(len(dmap_data)):
+        if abs(scan_mark[beam]) == 1:
+            no_scans += 1
+            beam_scan[beam] = no_scans
+        if scan_mark[beam] == 0:
+            beam_scan[beam] = no_scans
+    no_scans += 1
+    
+    return beam_scan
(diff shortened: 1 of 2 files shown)


pydarn/utils/scan.py
```diff
@@ -0,0 +1,34 @@
...
+This module is is used for sorting a given dmap_data list of dictionaries
+into scans
+"""
+
+import numpy as np
+from typing import List
+
+def build_scan(dmap_data: List[dict]):
+    """
+    Returns list of size equal tonumber of records in dmap_data, with scan number
```
carleyjmartin, line 13:
```suggestion
    Returns list of size equal to number of records in dmap_data, with scan number
```

pydarn/utils/scan.py
```diff
@@ -0,0 +1,34 @@
...
+    for each record
+
+    Parameters
+    ----------
+    dmap_data: List(dict)
+        list of records (dictionaries) representing dmap data
+    """ 
+    # Setup scans for easy locating
```
carleyjmartin, line 21:
```suggestion
    # Set up scans for easy locating
```

pydarn/utils/scan.py
```diff
@@ -0,0 +1,34 @@
...
+    scan_mark = [sub['scan'] for sub in dmap_data]
+    no_scans = 0
+    beam_scan = np.zeros((len(dmap_data)))
+    for beam in range(len(dmap_data)):
+        if abs(scan_mark[beam]) == 1:
+            no_scans += 1
+            beam_scan[beam] = no_scans
+        if scan_mark[beam] == 0:
+            beam_scan[beam] = no_scans
+    no_scans += 1
```
carleyjmartin, line 32: Is this line doing anything, you're already outside the for and returning beam_scan? Is it supposed to be inside the if?
```suggestion
            no_scans += 1
```

pydarn/utils/scan.py
```diff
@@ -0,0 +1,34 @@
+# Copyright (C) 2020 SuperDARN Canada, Universtiy of Saskatchewan
+# Author(s): Daniel Billett
+"""
+This module is is used for sorting a given dmap_data list of dictionaries
```
carleyjmartin, line 4:
```suggestion
This module is used for sorting a given dmap_data list of dictionaries
```

pydarn/utils/scan.py
billetd: Turns out it's actually redundant. Deleting.

billetd:
Committed all the changes, so happy for this to be merged if no other issues found. This will actually work with other scan modes if the scan flags are set correctly in file, now I think about it.

A seperate PR will come after this is merged which will add the functionality to generate a "scan array", i.e. a n_beams x n_gates array with the chosen data in it (pwr, vel, etc...).

pydarn/utils/scan.py
```diff
@@ -0,0 +1,33 @@
...
+
+def build_scan(dmap_data: List[dict]):
+    """
+    Returns list of size equal to number of records in dmap_data, with scan number
+    for each record
+
+    Parameters
+    ----------
+    dmap_data: List(dict)
+        list of records (dictionaries) representing dmap data
```
mts299, line 19: add returns field in

pydarn/utils/scan.py
```diff
@@ -0,0 +1,33 @@
...
+This module is used for sorting a given dmap_data list of dictionaries
+into scans
+"""
+
+import numpy as np
+from typing import List
+
+def build_scan(dmap_data: List[dict]):
+    """
+    Returns list of size equal to number of records in dmap_data, with scan number
```
mts299, line 13: What is the purpose of this function? Sorry, I think I lost track of something.

pydarn/utils/scan.py
```diff
@@ -0,0 +1,33 @@
...
+
+    Parameters
+    ----------
+    dmap_data: List(dict)
+        list of records (dictionaries) representing dmap data
+    """ 
+    # Set up scans for easy locating
+    # Makes a list of size (number of records), with the scan number for each
+    scan_mark = [sub['scan'] for sub in dmap_data]
+    no_scans = 0
```
mts299, line 24: Are no_scans supposed to be number of scans? 
maybe change to num_scans if so.

pydarn/utils/scan.py
```diff
@@ -0,0 +1,33 @@
...
+    beam_scan = np.zeros((len(dmap_data)))
+    for beam in range(len(dmap_data)):
+        if abs(scan_mark[beam]) == 1:
```
mts299, line 27: Add a comment about why abs and why 1 vs. 0 so someone who is not familiar with scan flags can understand this.

aburrell:
Just clocked this, and already had a similar function under development in a branch in pyDARNio.  Since this is a utility that is useful for more than plotting, perhaps we should have a discussion about where it should go.

mts299:
@aburrell I think the functionality will be a bit different as @billetd is making this more specifically in mind with fan plots. Correct me if I am wrong, but I think your utility will be broader for generally scientific reasons. 

With that said, let us keep this functionality for now and merge the PR once all comments are addressed @billetd and let you make more progress on fan plots instead or waiting for pyDARNio to catch up. If, in the end, we put it in pyDARNio, then we have to modify the code assuming the same functionality. Plus, I don't think this will be expressed as a primary user utility in pyDARN, so that the deprecation will be more silent. 

People want fan plots bad, so we need to keep this a priority for pyDARN as we are not dedicated to pyDARNio now.

aburrell:
> Correct me if I am wrong, but I think your utility will be broader for generally scientific reasons.

There are more functions, but there is also an essentially identical function.

> People want fan plots bad, so we need to keep this a priority for pyDARN as we are not dedicated to pyDARNio now.

Good enough for me!  Deprecation is easier than parallel development of dependencies.

mts299:
@billetd *poke* *poke*

billetd:
> @billetd _poke_ _poke_

Sorry, just got some revisions to do this week, then I'll get back on this.

pydarn/utils/scan.py
billetd: It's the current scan number used when iterating over all the records. I'll change to "current_scan"

billetd:
All done with this now. As soon as it's merged, can go back to fan plotting branch.

PR #77, opened 2020-10-01
prdarnio import

mts299:
# Scope 

To test the import of pydarnio as an option to the user if they wish to use with pyDARN. 

## Test 

`pip3 install .[io]` 

should see pydarnio installed in the output. 

Cannot use `setup.py` will need to test later with PyPi. 

## Draft 

draft for now as I need to code more to get the pydarnio API working within pyDARN.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -63,6 +63,6 @@ setup(
     # used to import the logging config file into pydarn.
     include_package_data=True,
     # setup_requires=['pyyaml', 'numpy', 'matplotlib', 'aacgmv2'],
-    install_requires=['pyyaml', 'numpy', 'matplotlib', 'aacgmv2', 'flake8']
-
+    install_requires=['pyyaml', 'numpy', 'matplotlib', 'aacgmv2', 'flake8'],
+    extras_require={'io':['pydarnio']},
 )
--- a/setup.py
+++ b/setup.py
@@ -62,10 +62,7 @@ setup(
     author="SuperDARN",
     # used to import the logging config file into pydarn.
     include_package_data=True,
-    setup_requires=['pyyaml', 'numpy', 'matplotlib', 'aacgmv2'],
-    # pyyaml library install
-    install_requires=['pyyaml', 'numpy', 'matplotlib', 'aacgmv2']
-    # commented out due to not implemented yet.
-    #ext_modules = [rstmodule]
+    # setup_requires=['pyyaml', 'numpy', 'matplotlib', 'aacgmv2'],
+    install_requires=['pyyaml', 'numpy', 'matplotlib', 'aacgmv2', 'flake8']
 
 )
--- a/setup.py
+++ b/setup.py
@@ -63,6 +63,5 @@ setup(
     # used to import the logging config file into pydarn.
     include_package_data=True,
     # setup_requires=['pyyaml', 'numpy', 'matplotlib', 'aacgmv2'],
-    install_requires=['pyyaml', 'numpy', 'matplotlib', 'aacgmv2', 'flake8'],
-    extras_require={'io':['pydarnio']},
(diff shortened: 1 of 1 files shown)


billetd:
I am very strongly of the opinion that installing IO should be the default, with the option to not install it.

mts299:
okay, any other thoughts? @carleyjmartin @mardet987

mts299:
not sure how to do the option of not installing .., can look into it

carleyjmartin:
I can't really see a reason why anyone would use pyDARN without the io utilities alongside it. So I agree that io should be installed with pyDARN as default. 

The main thing for me is that I'm not entirely sure what this means, will I need to go through any code already made and change all the io utils to be from pyDARNio, and import it too? This isn't a big issue as long as it stays the same in the future, but I'm not sure how intuitive that is to explain to a new student/scientist that some things have to be imported using pyDARNio and somethings have to be imported from pyDARN, I can imagine someone just asking why they can't just get it all from a pyDARN import. IMO anyway, which isn't that strong.

mts299:
if I make it a default in pyDARN I will be inheriting the io abilities and have a reading method in pyDARN for the API to work. It may not be the same reading name but you won't need to call pyDARNio.

mardet987:
I don't have the experience using this package standalone (without IO) - will users do this? I actually originally thought that it would have to be a dependency but now thinking about it I see that that's not necessary right now. 

Would you ever consider modifying your plotting functions to provide an argument for filename (not already read data), and to use pydarnio to read the file within the plotter? This maintains proper tracebacks with IO errors coming from pydarnIO, but might also make things easier for the user.

mts299:
@mardet987 thanks for the input, I have but sometimes users want to modify or put other data alongside it. So I like passing in the data similar to the API of matplotlib which pyDARN is essentially the extension of. 

I can see in the future though adding this if people really want it. 

Another reason to avoid is if you want to read multiple files, cat them together in proper order or a given order. It would be hard for a plotting algorithm to do that, which is something we do with our summary plots.

mardet987:
that makes sense. I like your plan of a reading method in pydarn that uses it so everything can be done from within pydarn but its still pretty modular that the IO is in pydarnIO. So I guess I support requiring the IO package and installing it by default.

mts299:
Done, switched it to default!

billetd:
In that case I'm happy with this as is, and think it should be merged quickly

billetd:
In that case I'm fine with this. I think it should be a priority to get this up and working.

carleyjmartin:
I just tested it with some of the old ACF code:
```
import pydarn
import pydarnio
from datetime import datetime

reader = pydarnio.SDarnRead("/Users/carley/Documents/data/20140105.1200.03.cly.rawacf")
data = reader.read_rawacf()
pydarn.ACF.plot_acfs(data, beam_num=15, gate_num=16,
                start_time=datetime(2014, 1, 5, 13, 30))
plt.show()
```
it works great when importing pydarnio and using that for SDarnRead, but doesn't work if I do

```
import pydarn
#import pydarnio
from datetime import datetime

reader = pydarn.SDarnRead("/Users/carley/Documents/data/20140105.1200.03.cly.rawacf")
data = reader.read_rawacf()
pydarn.ACF.plot_acfs(data, beam_num=15, gate_num=16,
                start_time=datetime(2014, 1, 5, 13, 30))
plt.show()
```

I get: 
``` 
AttributeError: module 'pydarn' has no attribute 'SDarnRead'
```
Is this the intended usage?
Or am I jumping the gun testing this now? :P

mts299:
io is removed remember lol

mts299:
see PR #78

mts299:
Closing this PR as the other encompasses everything since we made `pydarnio` a default.

Issue #103, opened 2021-02-23
Grid Plotting


mts299:
# New Feature Nottification

**Name**: *plot_grid*

**module**:  *grid.py* 

**package**: *plotting*

## Scope

This feature will plot grid files on a polar plot in magnetic coordinates. 
Will not plot on a geographic map or geo-coordinates. 

## Description

We will use a grid file to plot data on the polar plot from magnetic coordinates.  

### pyDARN Checklist 

-  [x] Does this fit within pyDARN's scope? 
-  [x] Is this a minor change i.e., new small utility function?
-  [ ] Is this a major change i.e., new plotting function?
-  [ ] Do you need help developing it? 
-  [ ] Have you created a pyDARN GitHub [project](https://github.com/SuperDARN/pydarn/projects) to show the tasks needed to be done?

PR #106, opened 2021-03-03
[EHN] plot Grid 

mts299:
# Scope 

New feature to plot grid plots on polar magnetic coordinates in MLT 

It does not plot multiple radar grid files. 

issue: #103 

## Approval

**Number of approvals:** 2, 1 reviewer, 1 tester 
However, I need opinions on the color bar 

## Test

```
import datetime as dt
import pydarn

#Read in fitACF file
grid_file = "20011220.0800.00.han.grid"

fanplot = pydarn.Grid.plot_grid(grid_data, start_time=dt.datetime(2001,12,20,8,30))
plt.show()
```

Figure_1

You can also test other parameters: `pwr` and `wdt`


pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+"""
+Grid plots, mapped to AACGM coordinates in a polar format
+"""
+
+import datetime as dt
+import matplotlib.pyplot as plt
+import numpy as np
+import cmocean
```
billetd, line 23: I ran into an issue where I needed to manually install 'cmocean' colormaps with pip before I could run code. Perhaps we should be sticking with a colormap that's already being used elsewhere? Any linear one would be okay I think, e.g. one used for power.

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+                Velocity value to be used for the reference vector, in m/s
+                Default: 300
+
+        See Also
+        --------
+        plot_fov - plots the field of view found in fan.py
+
+        Returns
+        -----------
+        if parameter is vector.vel.median:
```
billetd, line 122:
```suggestion
        If parameter is 'vel':
```

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+        thetas - list of data point longitudes in polar coordinates shifted
```
billetd, line 123:
```suggestion
        thetas - List of gridded data point magnetic local times (degrees)
```

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+        end_thetas - list of vector tails theta
```
billetd, line 124:
```suggestion
        end_thetas - List of magnetic local time end points used for vector plotting (degrees)
```

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+        rs - list of data point latitudes
```
billetd, line 125:
```suggestion
        rs - List of gridded data point radius' (AACGM latitude)
```

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+        end_rs - list of vector tails radius (latitude)
```
billetd, line 126:
```suggestion
        end_rs - List of radius end points for vector plotting (AACGM latitude)
```

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+        data - Line of Sight points of parameter chosen
```
billetd, line 127:
```suggestion
        data - List of magnitudes of line-of-sight velocity
```
The issue with returning the data when velocity is the parameter is that it's useless without `kvect` as well. There is no way to know whether the velocity is towards or away from the radar. I suggest also returning `kvect`.

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+        else:
+        thetas - list of data point longitudes in polar coordinates shifted
```
billetd, line 129:
```suggestion
        thetas - List of gridded data point magnetic local times (degrees)
```

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+        rs - list of data point latitudes
```
billetd, line 130:
```suggestion
        rs - List of gridded data point radius' (AACGM latitude)
```

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+        data - gridded vector points of the parameter chosen
```
billetd, line 131:
```suggestion
        data - List of data magnitudes plotted, for parameter chosen
```
If it's not velocity then it's not a vector.

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+            # Convert back to polar for plotting
+            end_rs = 90 - (np.sqrt(end_pos_x**2 + end_pos_y**2))
+            end_thetas = np.arctan2(end_pos_y, end_pos_x)
+
+            # Plot the vectors
+            for i in num_pts:
+                plt.polar([thetas[i], end_thetas[i]], [rs[i], end_rs[i]],
+                          c=cmap(norm(data[i])), linewidth=0.5)
+
+            # TODO: Add a velocity reference vector
```
billetd, line 234: Now I've thought about it, maybe we don't need a reference vector when there is also a colour bar. It would be needed however if the functionality is added to plot velocity vectors, but colour them by another parameter (like power). I could see that being useful, but might be a confusing plot.

setup.py
```diff
@@ -74,5 +74,5 @@ def run(self):
     include_package_data=True,
     # setup_requires=['pyyaml', 'numpy', 'matplotlib', 'aacgmv2'],
     install_requires=['pyyaml', 'numpy', 'matplotlib', 'aacgmv2',
-                      'pydarnio'],
+                      'pydarnio', 'cmocean'],
```
billetd, line 77: See earlier comment about not requiring more things than necessary. One of the matplotlib colormaps should be okay for our purpose.

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+            ticks = locator.tick_values(vmin=zmin, vmax=zmax)
+
+            cb = ax.figure.colorbar(mappable, ax=ax,
+                                    extend='both', ticks=ticks)
+
+            if colorbar_label != '':
+                cb.set_label(colorbar_label)
+
+        if title == '':
+            title = "{year} {month} {day} {start_hour}:{start_minute} -"\
```
billetd, line 249:
```suggestion
            title = "{year}-{month}-{day} {start_hour}:{start_minute} -"\
```

billetd (commented):
Something I want to do, maybe before this PR gets committed, is adding the functionality for fan plots to work on other axis that aren't polar. That change can then be incorporated into this grid plots. It should be done before we add too much plotting functionality and dig a deep hole! I'll try and get this done soon.

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+        -----------
+            dmap_data: List[dict]
+                Named list of dictionaries obtained from SDarn_read
+            record: int
+                record number to plot
+                default: 0
+            start_time: datetime.datetime
+                datetime object as the start time of the record to plot
+                if none then record will be used
+                default: none
```
carleyjmartin, line 68: Should you be able to use
```python
```
and it pick out the first record for you and plot that? if so it can't, I get an error with it trying to find dtime later on if start_time is not given (line153).

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+            cmap: matplotlib.cm
+                matplotlib colour map
+                https://matplotlib.org/tutorials/colors/colormaps.html
+                Default: Official pyDARN colour map for given parameter
+            zmin: int
+                The minimum parameter value for coloring
+                Default: {'pwr': [0], 'vel': [0], 'wdt': [0]}
+            zmax: int
+                The maximum parameter value for  coloring
+                Default: {'pwr': [50], 'vel': [1000], 'wdt': [250]}
```
carleyjmartin, line 96: Are these different defaults to other plots?

carleyjmartin (commented):
Tested using
```python
import datetime as dt
import pydarn

grid_file = "/Users/carley/Documents/data/grids/20150301.2001.00.inv.grid"

fanplot = pydarn.Grid.plot_grid(grid_data, start_time=dt.datetime(2015,3,1,20,30), parameter = 'vel')

plt.show()
```
The plot looks great!
No issues other than not being able to figure out if I can plot elevation angles or not? I don't know if that's a useful thing to have gridded or not. And I get the warning:
`295: UserWarning: FixedFormatter should only be used together with FixedLocator`


When we get to adding grid files together to show other radar data, I guess we would prefer the current colorbar over a diverging one and only abs values because it's not LOS anymore- it looks a lot like viridis reversed to me so I don't know if we need a dependency to cmocean just for this colormap?

pydarn/plotting/grid.py
carleyjmartin: I think having a reference vector as an option anyway would be good. Even if the default is to have it turned off.

pydarn/plotting/grid.py
billetd: It makes sense for the defaults to be different for `vel`, as the absolute magnitude is being plotted instead of a LOS value (which can be negative). But the `zmin` and `zmax`, for `pwr` and `wdt`, should probably be the same as the range time plots if they're not already.

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+
+            # Get the azimuths from the data
+            azm_v = dmap_data[record]['vector.kvect']
+
+            # Number of data points
+            num_pts = range(len(data))
+
+            # Angle to "rotate" each vector by to get into same reference frame
+            # Controlled by longitude, or "mltitude"
+            alpha = thetas
```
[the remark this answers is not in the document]
billetd: That's fair, but I suppose it was to distinguish `alpha` as the rotation parameter, which happens to also be the same as `thetas`.

pydarn/plotting/grid.py
billetd: It makes sense for the defaults to be different for `vel`, as the absolute magnitude is being plotted instead of a LOS value (which can be negative). But the `zmin` and `zmax`, for `pwr` and `wdt`, should probably be the same as the range time plots if they're not already.

pydarn/plotting/grid.py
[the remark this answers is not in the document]
carleyjmartin: Just make sure you're not doing anything to either one of the values later because depending on the type of copy you can end up changing both of them instead of making a copy and then only changing one.
I don't think you are though so it should be fine.

pydarn/plotting/grid.py
mts299: yea we, just the velocity for range time was misleading

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,261 @@
...
+        rs - list of data point latitudes
+        end_rs - list of vector tails radius (latitude)
+        data - Line of Sight points of parameter chosen
+        else:
+        thetas - list of data point longitudes in polar coordinates shifted
+        rs - list of data point latitudes
+        data - gridded vector points of the parameter chosen
+        """
+        # Short hand for the parameters in GRID files
+        if parameter == 'vel' or parameter == 'pwr' or parameter == 'wdt':
```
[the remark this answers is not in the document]
mts299: We used those ones because 'v'  was specific  to fitacf files  but  GRID files use `vel` `pwr` `wdt` and  MAP files so trying to be  consistent with the files parameters, not some other formed standard

pydarn/plotting/grid.py
mts299: Correct

pydarn/plotting/grid.py
mts299: fixed

pydarn/plotting/grid.py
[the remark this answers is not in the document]
mts299: because he is not storing anything in alpha it's not affecting thetas, but he could just call all `thetas` `alpha`. 

Why  `alpha`?

mts299:
>  No issues other than not being able to figure out if I can plot elevation angles or not? I don't know if that's a useful thing 

Elevation angles are not part of Grid files. I left a comment on why I  chose  `vel`, `pwr`, `wdt` since those are parameters in GRID  files.  

> And I get the warning:
295: UserWarning: FixedFormatter should only be used together with FixedLocator

Yea, I made an issue for that... not sure how to fix it yet 

>  I don't know if we need a dependency to cmocean just for this colormap?

is there a reason for not wanting another dependency normally when installing you wouldn't notice?

mts299:
>Something I want to do, maybe before this PR gets committed, is adding the functionality for fan plots to work on other axis that aren't polar. That change can then be incorporated into this grid plots. It should be done before we add too much plotting functionality and dig a deep hole! I'll try and get this done soon.

@billetd if you make this a function so any plotting method can use it, it won't dig a big hole. I would hate to hold up plotting methods for other things to get done.

pydarn/plotting/grid.py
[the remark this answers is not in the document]
carleyjmartin: Yeah I realized this after Dan talked me though what grid files actually were :P
As long as it's documented then that's all fine.

pydarn/plotting/grid.py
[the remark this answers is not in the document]
carleyjmartin: I think it's the letter alpha in most maths books for this rotation. But we do use alpha in another file (for fan plots now?)  in another way, so we might want to choose something else instead someone forgets and we end up mixing this value with transparency.

billetd:
> is there a reason for not wanting another dependency normally when installing you wouldn't notice?

Just to keep it as compact as possible really. Unless cmocean provided colormaps which some plotting routines couldn't do without, then it's just extra fluff.

billetd:
Something I just noticed is that `plot_grid` will add a few things which aren't in `plot_fan`, but probably should be. These are the plot title (date and time), and allowing the user to input a time rather than a scan number.

Devin-Huyghebaert:
Tested, looks good.  Only some minor issues.  I have discussed the first two with @mts299 already.

1. FOV does not extend based on input grid file.  Example fhw, with the figure below.
make_grid_fhw

Grid file generated from 20170729.0202.00.fhw.fitacf


2. Below issue was fixed with -xtd argument when generating the grid file.

parameter = 'pwr' or 'wdt' gives an error, parameter = 'vel' does not.

grid.py, line 194, in plot_grid
    data = dmap_data[record][parameter]
KeyError: 'vector.pwr.median'

grid.py, line 194, in plot_grid
    data = dmap_data[record][parameter]
KeyError: 'vector.wdt.median'

3. I believe the lower latitude bound for fan plots was set to 30 degrees?  Should this also be used for the grid plots?

- Devin

mts299:
@Devin-Huyghebaert 
1. Made a bug  report on it #116 as this is more of FOV bug than a grid bug 
2. added in the exception for that:
```
pydarn.exceptions.plot_exceptions.UnknownParameterError: The following parameter vector.pwr.median was not found in the data set. Please make sure you used -xtd if you are trying to plot pwd or wdt
```

3. Wait I thought we switched it.... well switched again!

billetd:
> Tested, looks good. Only some minor issues. I have discussed the first two with @mts299 already.
> 
> 1. FOV does not extend based on input grid file.  Example fhw, with the figure below.
>    make_grid_fhw
> 
> Grid file generated from 20170729.0202.00.fhw.fitacf
> 
> 1. Below issue was fixed with -xtd argument when generating the grid file.
> 
> parameter = 'pwr' or 'wdt' gives an error, parameter = 'vel' does not.
> 
> grid.py, line 194, in plot_grid
> data = dmap_data[record][parameter]
> KeyError: 'vector.pwr.median'
> 
> grid.py, line 194, in plot_grid
> data = dmap_data[record][parameter]
> KeyError: 'vector.wdt.median'
> 
> 1. I believe the lower latitude bound for fan plots was set to 30 degrees?  Should this also be used for the grid plots?
> 
> * Devin

billetd:
Just want to comment here as well as in PR #121 that the issue with FHW (and probably other radars too) isn't really a "bug", it is just that the default was originally to cut off at 75 range gates (as it was with fan plotting and FOV plotting). It clearly makes sense for the 75 cutoff to not be done here in `plot_grid', we should discuss whether to remove that default for the other plotting functions as well.

mts299:
Should be working with all the new changes: 

multiplot
grid1


``` python
import datetime as dt
import pydarn

#Read in fitACF file
stime = dt.datetime(2015, 3, 8, 14, 2)
#grid_file = "20150308.1400.03.cly.grid"
grid_file = "20150308.n.grd"

pydarn.Grid.plot_grid(grid_data, start_time=stime,
                      radar_label=True, fov_color='red',
                      fov_fill_color='grey')

plt.show()
```

billetd (approved):

pydarn/plotting/fan.py
```diff
@@ -180,7 +179,7 @@ def plot_fan(cls, dmap_data: List[dict], ax=None,
         # Plot FOV outline
         beam_corners_aacgm_lats, beam_corners_aacgm_lons, thetas, rs, ax = \
             cls.plot_fov(stid=dmap_data[0]['stid'], date=date,
-                         ranges=ranges, boundary=boundary,
+                         ranges=ranges, boundary=boundaRy,
```
carleyjmartin, line 182:
```suggestion
                         ranges=ranges, boundary=boundary,
```
not sure why, but it worked with the typo anyway for plotting FOV or grids, it breaks for fan plots

carleyjmartin (approved):
one teeny tiny typo

pydarn/plotting/grid.py
```diff
@@ -0,0 +1,294 @@
...
+                end_pos_x = start_pos_x + (vec_x / len_factor)
+                end_pos_y = start_pos_y + (vec_y / len_factor)
+
+                # Convert back to polar for plotting
+                end_rs = 90 - (np.sqrt(end_pos_x**2 + end_pos_y**2))
+                end_thetas = np.arctan2(end_pos_y, end_pos_x)
+
+                # Plot the vectors
+                for i in num_pts:
+                    plt.polar([thetas[i], end_thetas[i]], [rs[i], end_rs[i]],
```
carleyjmartin, line 262: ooo actually, do we need to change this to plt.plot with that issue with the new matplotlib?

carleyjmartin (commented):
oops sorry this occurred to me after the fact

pydarn/plotting/grid.py
carleyjmartin: as long as the ax object has that polar=True option set(which is read in from plot_fov I think anyway) just amending to plot should work

carleyjmartin, line 262:
```suggestion
                    plt.plot([thetas[i], end_thetas[i]], [rs[i], end_rs[i]],
```

sorry for the many notifications

carleyjmartin (approved):

PR #112, opened 2021-03-10
[EHN] Multi Grid plotting

mts299:
# Scope 

Just like  #106 I added so it can plot multiple radars values in a `GRD` file 

## Approval

**Number of approvals:** 2 - 1 reviewer, 1  tester

## Test

```python
import datetime as dt
import pydarn

#Read in fitACF file
grid_file = "20100305.n.grd"

plt.show()
```

To make  GRD  files you need RST:
```
combine_grd 20100305.inv.grid 20100305.pyk.grid 20100305.rkn.grid 20100305.sas.grid 20100305.wal.grid >  20100305.n.grd
```

I also have files if you need some

### DIFF
--- a/pydarn/plotting/grid.py
+++ b/pydarn/plotting/grid.py
@@ -257,12 +257,12 @@ class Grid():
             title = "{year}-{month}-{day} {start_hour}:{start_minute} -"\
                 " {end_hour}:{end_minute}"\
                     "".format(year=dtime.year,
-                              month=dtime.month,
-                              day=dtime.day,
-                              start_hour=dtime.hour,
-                              start_minute=dtime.minute,
-                              end_hour=dmap_data[record]['end.hour'],
-                              end_minute=dmap_data[record]['end.minute'])
+                              month=str(dtime.month).zfill(2),
+                              day=str(dtime.day).zfill(2),
+                              start_hour=str(dtime.hour).zfill(2),
+                              start_minute=str(dtime.minute).zfill(2),
+                              end_hour=str(dmap_data[record]['end.hour']).zfill(2),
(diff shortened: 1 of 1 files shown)


carleyjmartin (commented):
Tested with following code:
```python
import pydarn

grid_file = "/Users/carley/Documents/data/grids/20011220.8000.grid"

fanplot = pydarn.Grid.plot_grid(grid_data, parameter = 'pwr')

plt.show()
```

Produced following plots:

Screen Shot 2021-03-11 at 12 00 49 PM
Screen Shot 2021-03-11 at 11 57 33 AM

I think there might be a bug in the time for the title? It appears to be fine in the other single grid plot PR and I can't find where a title is made in the changes so it could have crept in somewhere else?
Otherwise, no issues I can find with the new for loop :D

billetd:
The time bug may have something to with specifying the time. It seems to work okay if you set `start_time`:

```python
grid_plot = pydarn.Grid.plot_grid(grid_data, start_time = dt.datetime(2001,12,20,9,30))
```

Which gives:
Screenshot 2021-03-11 at 12 30 25 PM

billetd:
Actually scratch my above comment. I think it's to do with when the minute value is under 10. Basically, no leading zero's are used in the plot title. So 9:01 shows as 9:1:

```python
grid_plot = pydarn.Grid.plot_grid(grid_data, start_time = dt.datetime(2001,12,20,9,0))
```
Which produces:
Screenshot 2021-03-11 at 12 36 29 PM

carleyjmartin (approved):

billetd (approved):

Issue #104, opened 2021-02-24
[BUG] UserWarning from Matplotlib


mts299:
# BUG 

When using `plot_fan` or `plot_fov` we get an `UserWarning` from matplotlib using `set_xtickslabels` but not setting the ticks beforehand. Because this is with a polar plot some thought needs to be given. 

## Priority

- [X] Minor - typo, incorrect naming, fails on specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

## Example of the bug 

```
294: UserWarning: FixedFormatter should only be used together with FixedLocator
287: MatplotlibDeprecationWarning: Adding an axes using the same arguments as a previous axes currently reuses the earlier instance.  In a future version, a new instance will always be created and returned.  Meanwhile, this warning can be suppressed, and the future behavior ensured, bypassing a unique label to each axes instance.
```

## Potential Bug Location 

https://github.com/SuperDARN/pydarn/blob/cf8d83e75253d53e8b31d3802c7f5d587183cba6/pydarn/plotting/fan.py#L191

PR #113, opened 2021-03-10
fixed the warning message on fan plots

mts299:
# Scope 

fixing the following `Userwarning`:
`UserWarning: FixedFormatter should only be used together with FixedLocator` 

**issue**: #104 

## Approval

**Number of approvals:** 1 this is a  1 liner fix

##  Test 

``` python
import datetime as dt
import pydarn

#Read in fitACF file
fitacf_file = "20100305.0001.00.inv.fitacf"
fitacf_data = pydarn.SuperDARNRead(fitacf_file).read_fitacf()

fanplot = pydarn.Fan.plot_fan(fitacf_data)
plt.show()
```

Figure_1

and no `userwarning`  message should appear

### DIFF
--- a/pydarn/__init__.py
+++ b/pydarn/__init__.py
@@ -42,4 +42,3 @@ from .plotting.rtp import RTP
 from .plotting.fan import Fan
 from .plotting.acf import ACF
 from .plotting.power import Power
-from .plotting.grid import Grid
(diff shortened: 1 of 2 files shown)


carleyjmartin:
The fix appears to be fine to me, but I think this was branched off from something where the Grid module was imported but it's not actually in develop as a module yet so I'm getting an import error:
```
File "/Users/carley/Documents/testing_code/venv-fanwarning/lib/python3.7/site-packages/pydarn/__init__.py", line 45, in <module>
    from .plotting.grid import Grid
ModuleNotFoundError: No module named 'pydarn.plotting.grid'
```

mts299:
huh fixed that  in develop not sure why it keeps occurring again 
try now?

carleyjmartin (approved):

Issue #114, opened 2021-03-11
Slant range calculation has incorrect "center" offset


ecbland:
# BUG 

When calculating the slant range in `gate2slant()`, the center=0 and center=1 cases are flipped.

## Priority

- [ ] Minor - typo, incorrect naming, fails on specific uncommon situation
- [X] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

## Example of the bug 

This bug was found by looking at the source code (see below), so I don't have any specific test code to demonstrate it.

## Potential Bug Location 

https://github.com/SuperDARN/pydarn/blob/452f36817794bb28b8819163f1dd837e2ecc824f/pydarn/utils/conversions.py#L114-L120

Compare to the [equivalent calculation in RST](https://github.com/SuperDARN/rst/blob/b667628c85580b9bbcff6b167345a33a5d0c1c56/codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c#L348-L353): 
```
    /* If not calculating center position of range-beam cell then calculate
     * position of near-left corner instead */
    if (center==0) {
        bm_edge=-pos->bmsep*0.5;
        range_edge=-0.5*rsep*20/3;
    }
```

## Potential Solution(s)

Should be a simple swap:
```
if center:
        range_offset = 0.0
else:
        # 0.5 offset to the edge of the range gate
        range_offset = -0.5 * record['rsep']
```

PR #115, opened 2021-03-11
fixed slant range centre bug

mts299:
# Scope 

Swapping  the center values to reflect the proper center of range gates for the slant range calculations

**issue**: #114

## Approval

**Number of approvals:** 1 approval simple fix 

## Test 

``` python
import pydarn
import pydarnio



fitacf_file = './data/20100305.1401.00.rkn.fitacf'
pydarn.RTP.plot_summary(fitacf_data, beam_num=15)
plt.show()
```

Figure_1


carleyjmartin:
I'm still getting the `ModuleNotFoundError: No module named 'pydarn.plotting.grid'` from the old develop branch. 

But removing that line from `__init__.py`, and using this code:
```python
import pydarn

#Load in dmap file
#Plot and show
#a = pydarn.RTP.plot_range_time(fitacf_data, beam_num=4, slant=True)
a = pydarn.RTP.plot_summary(fitacf_data, beam_num=2)
plt.show()
```
I think it might be causing a slight offset in the y-axis, it's barely noticeable in the summary plots:
Screen Shot 2021-03-19 at 1 46 09 PM
But on a RT plot you can see it clearly:
Screen Shot 2021-03-19 at 1 45 39 PM

I'm happy to look into it and see if I can't find the solution

mts299:
merged new `develop` in so should be fixed now

pydarn/utils/conversions.py
```diff
@@ -114,11 +114,9 @@ def gate2slant(record, nrang, center=True):
     # Range offset
     # If center is true, calculate at the center
     if center:
-        # 0.5 off set to the centre of the range gate instead of edge
-        range_offset = -0.5 * record['rsep']
-    else:
         range_offset = 0.0
-
+    else:
+        range_offset = -0.5 * record['rsep']
```
carleyjmartin, line 119:
```suggestion
        range_offset = 0.5 * record['rsep']
```
Thinking about this, and definitely get a second opinion on this because it's friday afternoon and I could just be suffering from a case of fridays, if it's not at the center of the range gate, then do we want to add the range offset? So are we adding the range offset from the edge of the range gate thats closest to the radar (inner) or are we removing it from the outer edge of the range gate? @billetd @Devin-Huyghebaert

carleyjmartin: Okay, so I think I'm just not understanding what the the center option is for. So it's default true, which means that the data is plotted at the center of the range-gate, if the user gives center=False then it switches to plotting at the start of the range gate, for which a - makes sense.

carleyjmartin (approved):

ecbland:
Do we need to add `center=FALSE` to the slant range calculation in `rtp.py`? The _edge_ of the first range gate should be at 180km.

https://github.com/SuperDARN/pydarn/blob/8ee9ecf0fc9184034a56c5961c0d8089c400a39f/pydarn/plotting/rtp.py#L340

I am a bit confused by the y-axis though, which @carleyjmartin mentioned above. It makes it difficult to check whether this is behaving correctly.

carleyjmartin:
I think the plots above in my review, the lowest number (180 in that case) is at the edge of the range gate, or at least it appears to be. But I guess we don't know if 180 is the edge value or the center value when it's plotting it. My guess is that gate2slant is returning the middle value of the range gate, but the rtp plotting is assuming that is the edge and plotting that value as the edge, so your fix of `center=False` would be the right way to go. 

I can play around with this today and plot some numbers out to definitively see what the plotting function is plotting and where the range gates are located in slant range.

There is a fix now merged to develop for the y-axis, but I don't think it would help for this purpose anyway.

carleyjmartin:
The plotting function plots from the position it's given, so it's assuming it's being given the edge of the range gate.
I believe that you do want `center=False` in the gate2slant function call, I guess this is only tangentially related to this PR so I'll make a new issue and sort it out there.

mts299:
@ecbland merged `develop` in which should update the ticks issue and we can merge again once #121 back in as well

pydarn/utils/conversions.py
```diff
@@ -114,11 +114,9 @@ def gate2slant(record, nrang, center=True):
     # Range offset
     # If center is true, calculate at the center
     if center:
-        # 0.5 off set to the centre of the range gate instead of edge
-        range_offset = -0.5 * record['rsep']
-    else:
         range_offset = 0.0
```
carleyjmartin, line 117:
```suggestion
        range_offset = 0.5 * record['rsep']
```

pydarn/utils/conversions.py
carleyjmartin:
```suggestion
        range_offset = 0.0
```

carleyjmartin (commented):
These last two suggestions are based off the discussion at the last pyDARN meeting. Using the assumption that the 'frang' value is defined as the **front edge of the range gate**. 
We know this assumption is not 100% - but within the remit of pyDARN this is a 'fix now, let higher up fix the definition of frang, and depending on the outcome, fix at that later time'.
This in conjunction with #134 completes the slant range fixes.

pydarn/utils/conversions.py
```diff
@@ -114,11 +114,9 @@ def gate2slant(record, nrang, center=True):
     # Range offset
     # If center is true, calculate at the center
     if center:
-        # 0.5 off set to the centre of the range gate instead of edge
-        range_offset = -0.5 * record['rsep']
+        range_offset = 0.5 * record['rsep']
     else:
-        range_offset = 0.0
-
+        range_offset = 0.5 * record['rsep']
```
carleyjmartin, line 119:
```suggestion
        range_offset = 0.0
```

Issue #116, opened 2021-03-15
FOV limited to 75 range gates by default


mts299:
# BUG 

FOV only shows the first 75 range gates which maybe not be correct for radars with a large FOV. 

## Priority

- [ ] Minor - typo, incorrect naming, fails on specific uncommon situation
- [x] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

## Example of the bug 

![Figure_1](https://user-images.githubusercontent.com/6520530/111198052-c7d90d00-8584-11eb-81c3-4a4beacbf72c.png)

### Data Location

`make_grid` with FHW data or any known radar with more than 75 Rage gates. 

## Potential Bug Location 

*Please note this is not required.*
*If you can link to some code in pyDARN where you think the bug is coming from this will help the developer fix it sooner.*

## Potential Solution(s)

https://github.com/SuperDARN/pydarn/blob/a61a5e37cd7a50e1c40629866fbe0564d1433222/pydarn/plotting/fan.py#L50

Change the default and get the range gate value from the hardware files.

mts299:
@billetd @carleyjmartin @Devin-Huyghebaert @ecbland here is a purposed solution to this problem:
In the `SuperDARNRadars` class that has all the radar info, I add `45_range` and `15_range` on to the radars info that will hold the max range gate for both 45 and 15 km modes then we can use that number (45km) to generate the FOV plots when 75 gates are not quite good enough like for `plot_grid`? 

We can get the range gate values from RAWACF/FITACF files since Hardware files seem to be non-standardized right now.

carleyjmartin:
Just so I'm understanding right: I want to plot a FOV plot, the FOV will be plotted using the number of range gates dictated by the data file I'm reading in. If I'm not reading in a data file, it grabs the number of range gates from `45_range` in `SuperDARNRadars`?
If so, that's good with me.

mts299:
No data file is used for a FOV plot originally I was hoping the Hardware file would be used but `gates` is a bit ambiguous in meaning right now (DSWG issue)

Yes, it would grab from `45_range`. Another reason to not grab from the data file is that we don't have FOV for 15 km mode so we wouldn't be able to plot 225 range gates and end up in the same spot. For now, we can say `plot_fan` and `pllot_fov` assume 45km mode as it is the most common mode.

[...]

PR #121, opened 2021-03-15
fixed the gate range default

mts299:
# Scope 

Resolving the FOV being too small for grid plotting 

**issue**: #116

## Approval

**Number of approvals:** 1 approval easy fix

# Test

```
import pydarn

#Read in fitACF file
grid_file = "20170729.0202.00.fhw.grid"

plt.show()
```

Figure_1

### DIFF
--- a/pydarn/plotting/fan.py
+++ b/pydarn/plotting/fan.py
@@ -274,7 +274,7 @@ class Fan():
             radar_fov(stid, coords='aacgm', date=dtime)
         fan_shape = beam_corners_aacgm_lons.shape
         if ranges == []:
-            ranges = [0, fan_shape[-1]]
+            ranges = [0, fan_shape[0]]
         # Work out shift due in MLT
         beam_corners_mlts = np.zeros((fan_shape[0], fan_shape[1]))
         mltshift = beam_corners_aacgm_lons[0, 0] - \
--- a/pydarn/plotting/fan.py
+++ b/pydarn/plotting/fan.py
@@ -295,6 +295,9 @@ class Fan():
             else:
                 ax.set_ylim(-90, -abs(lowlat))
                 ax.set_yticks(np.arange(-abs(lowlat), -90, -10))
+            ax.set_xticks([0, np.radians(45), np.radians(90), np.radians(135),
+                           np.radians(180), np.radians(225), np.radians(270),
+                           np.radians(315)])
             ax.set_xticklabels(['00', '', '06', '', '12', '', '18', ''])
             ax.set_theta_zero_location("S")
 
--- a/pydarn/plotting/fan.py
+++ b/pydarn/plotting/fan.py
@@ -27,7 +27,7 @@ from typing import List
 # Third party libraries
 import aacgmv2
 
-from pydarn import PyDARNColormaps, build_scan, radar_fov, SuperDARNRadars
+from pydarn import PyDARNColormaps, build_scan, radar_fov
 
 
 class Fan():
--- a/pydarn/plotting/fan.py
+++ b/pydarn/plotting/fan.py
@@ -270,13 +270,11 @@ class Fan():
             rs - radius polar coordinates
         """
         # Get radar beam/gate locations
-        if ranges == []:
-            max_ranges = SuperDARNRadars.radars[stid].hardware_info.gates
-            ranges = [0, max_ranges]
         beam_corners_aacgm_lats, beam_corners_aacgm_lons = \
             radar_fov(stid, coords='aacgm', date=dtime)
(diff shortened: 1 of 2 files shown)


Devin-Huyghebaert:
Tested with:

> 
> import matplotlib.pyplot as plt 
> import datetime as dt
> import pydarn
> 
> #Read in fitACF file
> grid_file = "data/20150318.0002.00.sas.grid"
> grid_data = pydarn.SuperDARNRead(grid_file).read_grid()
> 
> fanplot = pydarn.Grid.plot_grid(grid_data, start_time=dt.datetime(2015,3,18,0,30),parameter='vel')
> plt.show()

Gave error:

> grid.py, line 160, in plot_grid
>     _, aacgm_lons, _, _, ax = Fan.plot_fov(stid, dtime,
> fan.py, line 308, in plot_fov
>     plt.polar(thetas[ranges[1] - 1, 0:thetas.shape[1] - 1],
> IndexError: index 224 is out of bounds for axis 0 with size 76

Worked with fhw (20170729.0002.00.fhw.grid) and mcm (20161124.0001.00.mcm.a.grid) files I tried though.

mts299:
Ha you beat me but yes I just found this bug as well. fixing it now.

billetd:
I would like this to be put on hold for now, as it seems like a bandage over the actual issue. This seems to be that the radar FOV files themselves have a mis-match with the hardware files.

Also, in the example in the main comment, FHW is used which goes up to 110 range gates in both the HDW and radar FOV files. Was the issue simply that it defaulted to only showing 75 ranges in `plot_fov`?

pydarn/plotting/fan.py
```diff
@@ -268,6 +270,9 @@ def plot_fov(cls, stid: str, dtime: dt.datetime, ax=None,
             rs - radius polar coordinates
         """
         # Get radar beam/gate locations
+        if ranges == []:
+            max_ranges = SuperDARNRadars.radars[stid].hardware_info.gates
+            ranges = [0, max_ranges]
```
billetd, line 275: This part (L273-275) is unnecessary. By involving the hardware files, it allows the possibility for conflicts with the FOV files. Possible change suggested below.

pydarn/plotting/fan.py
```diff
@@ -268,6 +270,9 @@ def plot_fov(cls, stid: str, dtime: dt.datetime, ax=None,
...
         beam_corners_aacgm_lats, beam_corners_aacgm_lons = \
             radar_fov(stid, coords='aacgm', date=dtime)
         fan_shape = beam_corners_aacgm_lons.shape
```
billetd, line 278:
```suggestion
        fan_shape = beam_corners_aacgm_lons.shape
        
        if ranges == []
            ranges = [0, fan_shape[1]]
```

billetd (changes requested):

pydarn/plotting/fan.py
mts299: Well, that is rude? It is when you limit the FOV to being smaller than it really it is.

pydarn/plotting/fan.py
mts299: I will use this bandage to replace the actual necessary solution where the radar FOV files need to be consistent with the hardware files.

mts299:
So looks like this issue might be with the hardware files having 225 for our few radars soooo we will wait on discussing this with Kevin.

mts299:
@billetd switched to your solution as mine was unnecessary and just a ban aid. Sorry. 

You can take the lead from here and other problems and defaults in the future.

pydarn/plotting/fan.py
```diff
@@ -27,7 +27,7 @@
 # Third party libraries
 import aacgmv2
 
-from pydarn import PyDARNColormaps, build_scan, radar_fov
+from pydarn import PyDARNColormaps, build_scan, radar_fov, SuperDARNRadars
```
carleyjmartin, line 30:
```suggestion
from pydarn import PyDARNColormaps, build_scan, radar_fov
```
Imported but not used

carleyjmartin (approved):
Tested with:
```python
import datetime as dt
import pydarn

#grid_file = "/Users/carley/Documents/data/grids/20011220.8000.grid"
grid_file = "/Users/carley/Documents/data/grids/20011220.0800.00.han.grid"

fanplot = pydarn.Grid.plot_grid(grid_data, parameter = 'pwr')

plt.show()
```
for grid plot examples and tested on some old fan plotting code just to check and no issues now Dan fixed the typo in fan_shape. Just one suggestion on an import not used.

mts299:
@billetd is this good to merge?

billetd (approved):

Issue #126, opened 2021-03-19
Discrepancy between major and minor ticks in range-time plots


carleyjmartin:
# BUG 

When plotting range-time plots or summary plots using the `slant=True` option the major ticks are shown at 200km starting at ~180km. Minor ticks are set to 100km, assuming that they are starting from the nearest '100' which in this case is at 200 km, +20km from the major ticks.

## Priority

- [ ] Minor - typo, incorrect naming, fails on specific uncommon situation
- [x] Major - misrepresentation of data, failure in a common situation (technically, but they're not labelled and are pointing to the full rounded hundreds so it's not a horrific misrepresentation of the data)
- [ ] Urgent - failure in usage, large misrepresentation

## Example of the bug 

```python
import matplotlib.pyplot as plt 
import pydarn

#Load in dmap file
file = "/Users/carley/Documents/data/20161123.1601.00.mcm.a.fitacf"
SDarn_read = pydarn.SuperDARNRead(file)
fitacf_data = SDarn_read.read_fitacf()
#Plot and show
a = pydarn.RTP.plot_range_time(fitacf_data, beam_num=4, slant=True)
plt.show()
```
![Screen Shot 2021-03-19 at 2 16 06 PM](https://user-images.githubusercontent.com/60905856/111837752-f6752180-88bd-11eb-8c66-53b865a5d37d.png)

### Attempts

*What have you tried already to fix it?*

*What have you tried to get around it?*

### Data Location

Any fitacf file should recreate this bug in range-time plots. 

## Potential Bug Location 

Line 402 and 417 in rtp.py
```python
        if slant:
            ax.yaxis.set_ticks(np.arange(ymin, ymax+1, yspacing))
        else:
            ax.yaxis.set_ticks(np.arange(ymin, ymax+1, (ymax)/5))

.......
        if slant:
            ax.yaxis.set_minor_locator(ticker.MultipleLocator(100))
        else:
            ax.yaxis.set_minor_locator(ticker.MultipleLocator(5))
```
## Potential Solution(s)

Possibly need to set the base, or bottom tick for both to be the same?

## Extra Notes

No issue with range gates for the plots. 
While we're playing with ticks, it might be nice to have them both round off at the hundreds?

PR #127, opened 2021-03-19
FIX/Offset between major and minor ticks using slant range in RTPs

carleyjmartin:
# Scope 

To fix issue #126 

Fix for the offset between the major and minor ticks when using the slant=True option for RTP and subsequently summary plots. The minor ticks are now set to be in the middle of the major ticks. Previous to this, the ticks were set at every 100, so they may look more sparse on the summary plots as there would have been 4 minor ticks between each major tick, and now there is just 1 minor tick.
I have also included a quick change to the major ticks so that they show up on the nearest hundred and increase each 200 for RTP (still increase at 500 on summary plots but are rounded to nearest hundred too, possible future change to make summary plots put ticks on the nearest 500th unit?), mainly for neatness and ease of the user to read off the axis. 

## Approval

**Number of approvals:** 2 

  - copyrighted (if applicable)

Not modified the header of rtp.py as I'm not sure if it's still in the old style or not?

Example new RT plot axis:
Screen Shot 2021-03-19 at 4 16 21 PM
Example new summary plot axis:
Screen Shot 2021-03-19 at 4 15 54 PM
Above plots made with:
```python
import pydarn

#Load in dmap file
#Plot and show
a = pydarn.RTP.plot_range_time(fitacf_data, beam_num=4, slant=True)
#a = pydarn.RTP.plot_summary(fitacf_data, beam_num=2, slant=True)
plt.show()
```

### DIFF
--- a/pydarn/plotting/rtp.py
+++ b/pydarn/plotting/rtp.py
@@ -399,7 +399,8 @@ class RTP():
         ax.set_ylim(ymin, ymax)
 
         if slant:
-            ax.yaxis.set_ticks(np.arange(np.ceil(ymin / 100.0) * 100, ymax+1, yspacing))
+            ax.yaxis.set_ticks(np.arange(np.ceil(ymin/100.0)*100,
+                                         ymax+1, yspacing))
         else:
             ax.yaxis.set_ticks(np.arange(ymin, ymax+1, (ymax)/5))
 
--- a/pydarn/plotting/rtp.py
+++ b/pydarn/plotting/rtp.py
@@ -415,7 +415,7 @@ class RTP():
             tick_interval = 1
         ax.xaxis.set_minor_locator(dates.MinuteLocator(interval=tick_interval))
         if slant:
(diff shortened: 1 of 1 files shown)


pydarn/plotting/rtp.py
```diff
@@ -399,7 +399,7 @@ def plot_range_time(cls, dmap_data: List[dict], parameter: str = 'v',
         ax.set_ylim(ymin, ymax)
 
         if slant:
-            ax.yaxis.set_ticks(np.arange(ymin, ymax+1, yspacing))
+            ax.yaxis.set_ticks(np.arange(np.ceil(ymin / 100.0) * 100, ymax+1, yspacing))
```
carleyjmartin, line 402:
```suggestion
            ax.yaxis.set_ticks(np.arange(np.ceil(ymin/100.0)*100, 
                                                            ymax+1, yspacing))
```
Not entirely sure how indented that needs to be

Devin-Huyghebaert (approved):
Tested.  No issues found.

PR #134, opened 2021-03-23
[FIX] Plot RTPs from edge of range gate not center value

### LINKED ISSUE #132: Slant range RTP plots edge of range gate at center value
# BUG 

When plotting range-time plots with `slant=True`, the plot will plot the value give from the function `gate2slant` as the lower edge of the range gate. Currently `gate2slant` is returning the central value of the range gate so the plots are slightly off. 

## Priority

- [ ] Minor - typo, incorrect naming, fails on specific uncommon situation

## Example of the bug 
```python
import pydarn

#Load in dmap file
#Plot FOV and show
a = pydarn.RTP.plot_range_time(fitacf_data, beam_num=4, slant=True)
#a = pydarn.RTP.plot_summary(fitacf_data, beam_num=2, slant=True)
plt.show()
```

### Attempts

*What have you tried already to fix it?*
The above code plots the lowest range gate boundary at 180, it should be 180-22.5 which it does when you add `center=False` to the `gate2slant` function call. 
*What have you tried to get around it?*
NA

## Potential Bug Location 

L340 in rtp.py

## Potential Solution(s)

Add `center=False` to function call.

carleyjmartin:
# Scope 

Amending so the data is plotted from the inner edge of the range gate not the center of the range gate when converted to slant range. 

To fix #132 

## Approval

**Number of approvals:** 2 (only a one line fix so not hard)

## Test
```python
import pydarn
import logging

#Load in dmap file
file = ".fitacf" # any fitacf file here
#Plot FOV and show
a = pydarn.RTP.plot_range_time(fitacf_data, beam_num=4, slant=True)
#a = pydarn.RTP.plot_summary(fitacf_data, beam_num=2, slant=True)
plt.show()
```
Should notice that it begins plotting ~22.5km closer than it did before if on a 45km mode. (Rounded to int on axis though)

### DIFF
--- a/pydarn/plotting/rtp.py
+++ b/pydarn/plotting/rtp.py
@@ -337,7 +337,7 @@ class RTP():
                                      end_time=end_time,
                                      opt_beam_num=cls.dmap_data[0]['bmnum'])
         if slant:
-            y = gate2slant(cls.dmap_data[0], y_max)
+            y = gate2slant(cls.dmap_data[0], y_max, center=False)
         time_axis, y_axis = np.meshgrid(x, y)
         z_data = np.ma.masked_where(np.isnan(z.T), z.T)
         Default = {'noise.sky': (1e0, 1e5),


mts299:
Tested it and it works 

Figure_1

Lett me know if I should merge since its a simple fix

mts299:
One thing I barely can see the difference between `develop` but see a difference in `fix/slant_range_center`

mts299 (approved):

carleyjmartin:
Develop now has the axis fix merged in, so it will show probably 200km as the first tick regardless so you won't see the exact lower bound value on the axis, so you won't notice that 22.5km difference, but in the other branch it still plots the lowest value of slant range on the axis so it should change by 22/23 km on that axis.

mts299:
@ecbland does this look correct?

ecbland:
I'm getting confused here. Using RST's `time_plot` routine, the edge of the first range gate is at 180km (see below). On this branch in pydarn the first range gate starts at ~165km.

Would it make sense to merge #115 and then test this again?

```
# RST code
time_plot -png -v -ex 02:00 -st 06:00 -km -erang 630 -ymajor 45 -b 5 -vmin -500 -vmax 500 20150305.0601.00.inv.fitacf > plot.png
```
plot

carleyjmartin:
When plotting with pyDARN for Inuvik beam 5 on 05-03-2015 I get the lowest gate boundary plotted at (Sorry I don't get 165km, but I do get 165 km for RKN on the same day which I think is the example in the other PR?):
- Current master branch (no center fixes): 142 km 
- Develop without either fixes: 157.5 km (Don't know why this is different to master?)
- With ONLY center=False fix (this branch): 180 km
- With ONLY center bug fixed (#115 ): 180 km
- With center=False AND center bug fixes: 157.5 km

I don't have RST installed to test on there. My head is telling me that both of the fixes make sense, but essentially they're cancelling each other out in testing. I have no prior knowledge of RST, but I'm assuming RST plots the lowest bound at the correct lowest edge of the range gate (i.e. RST doesn't have either of these center bugs, and 180 km is the true value of the boundary we should aim for?)

The logical outcomes here then are:
- One of these fixes isn't needed? But for the life of me I couldn't tell you which one. 
- RST also has a centering issue for plotting. (Not likely but I'm covering bases)

carleyjmartin:
*Sorry, this turned into a lot of questions*

Just testing with different branches for pyDARN for INV beam 5 on 05-03-2015 between 06-08UT:
- Current develop (no fixes): 142.5 km
- With ONLY center=false fix (this branch): 165 km
- With ONLY center bug fixed (#115): 165 km
- With center bug AND center=False: 142.5 km

The two fixes are unfortunately cancelling each other out when testing. So I'm thinking one of them is not right. 

**Problem 1: Discrepancy between RST and pyDARN for first gate distance**
For this example `frang=180km`
I know literally 0 about RST, can we assume that 180 km is the absolutely true value we should be aiming for?
How does RST calculate slant range/ is it a different way than in pyDARN?

In pyDARN (pseudocode) we're calculating the distance to the first range gate as (added full code to bottom):
```
lag_first = frang * (2 / c)
sample_separation = rsep * (2 / c)
slant_range = (lag_first - risetime + gate#*sample_separation) * (2 / c)
```
So our first range gate is unlikely to be exactly 180 km... does RST assume the first range gate in slant range is at the value given by `frang`?

**Problem 2: `frang` definition and outcome in pyDARN for these PRs**
Is the `frang` value in a FITACF file the distance to the edge of the first range gate, or distance to the center of the range gate? The documentation just says 'Distance to the first range gate'.

If it is the edge of the first range gate, then the center=False (the value wanted/given by frange is not the center of the range gate) and then we don't want any modifiers on the `range_offset`, hence the fix for #115 #114 is not needed AND then the `-0.5 * record['rsep']` should be `+0.5 * record['rsep']` to bump it to the center of the first range gate. 

If the `frang` value gives the center of the first range gate, then center=True, we don't need this PR fix, and we want to `-0.5 * record['rsep']` the value given (i.e. we want #115 and #114 to be merged and not this branch)(and then RST has the center bug too).

This gave me new brain wrinkles. Please go to town on any assumptions I'm making. 
Less important side note: I think this is also a definition problem for what `center=True` is defined as, in here I've assumed that `center=True` = we want the slant range value that is the center of the range gate and `center=False`= we want the slant range value that is the edge of the range gate. The alternative, we could be thinking `center=True` = this function is being given the value at the center of the range gate, the function always wants to output the edge so we need to deal with that in the logic (-0.5 rsep needed), and `center=False` = this function has been given the edge value of the range gate, do nothing (0.0)

Appendix:
```python
speed_of_light = 0.3  # TODO: should this be more accurate?
    distance_factor = 2.0
    lag_first = record['frang'] * distance_factor / speed_of_light
    print(record['frang'])
    # sample separation in microseconds
    sample_sep = record['rsep'] * distance_factor / speed_of_light
    # Range offset
    # If center is true, calculate at the center
    if center:
        # 0.5 off set to the centre of the range gate instead of edge
        range_offset = 0.0   OR  -0.5 * record['rsep']
    else:
        range_offset = -0.5 * record['rsep']   OR 0.0

    # Now calculate slant range in km
    slant_ranges = np.zeros(nrang+1)
    for gate in range(nrang+1):
        slant_ranges[gate] = (lag_first - record['rxrise'] +
                              gate * sample_sep) * speed_of_light /\
                distance_factor + range_offset
    return slant_ranges
```

mts299:
Here is where RST calculates slant range: https://github.com/SuperDARN/rst/blob/0aa1fffed4cc48c1eb6372dfc9effa688af95624/codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c

carleyjmartin:
> Here is where RST calculates slant range: https://github.com/SuperDARN/rst/blob/0aa1fffed4cc48c1eb6372dfc9effa688af95624/codebase/superdarn/src.lib/tk/rpos.1.7/src/cnvtcoord.c

Brilliant thank you.

Okay, so I don't read C so this is what I get for the equation in RST separating out some values to see the difference in pseudocode, correct me if I'm wrong:
```
slant_range = (lagfr - rxrise + (range_gate - 1) * smsep + range_edge) * c/distance_factor
```
for pyDARN we have:
```
slant_range = (lagfr - rxrise + range_gate * smsep) * c/distance_factor + range_offset
```

With frng=180 and rsep=45...
Assuming rxrise=0.0 and the gate is 0 and range_edge=0, the RST equation gives us 180 km 👍 
With the same assumptions and range_offset=0, the pyDARN equation also gets 180 km - not 165 km 👎 

Even if we're accidentally not -1 off range_gate then that would still give 225 km - not 165 km. 

This is a conundrum, but we should definitely be expecting to get 180 km for our first range gate, I'll look more into this on Monday morning!

mts299:
Changed the base so we keep all slant range fixes together in case they affect one another.

carleyjmartin:
@ecbland 
On problem 1: It appears that the discrepancy between 165km and 180km is related to this old PR in RST: https://github.com/SuperDARN/rst/issues/191

RST is using a hardcoded rxrise value of 0.0 in the example above (assuming because the output in 180km) from the hardware files, and pyDARN is reading in a rxrise time of 100us from teh FITACF file, giving that -15km difference. 

The solution here would be to switch to get the rxrise value from hdw. BUT also has all the issues and caveats as the above linked issue in RST. Or ignore rxrise when plotting for now, which is a little safer?

On problem 2: Frang is most likely the edge of the first range gate.

mts299:
Added it to the telecon

carleyjmartin:
On problems 2&3:

Looking through davitpy, the boolean center value is defined as: 'whether or not to compute the slant range in the center of the gate rather than at the edge (default=True)' L272: https://github.com/vtsuperdarn/davitpy/blob/master/davitpy/pydarn/radar/radFov.py
Which for our RTP if we take the same definition, we want `center=False`, so we want this fix. 

Which means that we want the function in #115 to follow the logic:
``` 
if center=False: 
    range_offset = 0.0
else if center=True:
   range_offset = (+)0.5*rsep
```
All assuming `frang` is the inner edge of the first range gate.  
I don't know why in davitpy the function is -ve and the other way around unless davitpy is assuming that the frang value it's reading in is not the edge of the range gate? I also don't really understand why the default is true, as it would be used less often than false, but that might be for another reason in davitpy. 

Either way, this is only one way, there are many others so like I said in the above comment, it's probably something much easier discussed and understood by talking about it (and just saw the above comment, thanks @mts299!) . I'm just documenting on comments here so we have a record thats not my scribblings :)

mts299:
Thank you @carleyjmartin for documenting all this and @Devin-Huyghebaert for looking further into all this

asreimer:
I haven't had time to go through all of the discussion here in as much detail as it deserves. Sorry, I'm short on time...

But I think this is helpful context: The `center=True` stuff in `davitpy` was intended purely for building a grid of points that you could use to plot the data. For example, `pcolormesh` in `matplotlib` wouldn't plot a coloured square **centered** on the range gate unless we built a grid with start and end point that was +/- half a range gate from the slant range of the range gate.

carleyjmartin:
After discussion at the pyDARN meeting today. This in conjunction with #115 will 'fix' the slant range until now. Assuming that a center=False flag means that you want the function to return the edge of the range gate, center=True means that you want the function to return the central value of the range gate. This means that we can use this function for other reasons too. 

This is ready to merge IMO.

Issue #133, opened 2021-03-23
Coodinate System  Enum


mts299:
# New Feature

**Name**: `class  Coord`

**module**:  *coordinates.py* 

**package**: *utils/*

## Scope

This coordinate enum will represent the possible coordinate systems pyDARN can offer to the plotting methods. 

The class will just allow the user to pick a constant name already set avoiding typos and miss understandings on what is implemented. 

## Description

This feature will allow easier understanding for what coordinate systems are in place as well as allow flexibility to add other in the future. 

### pyDARN Checklist 

-  [x] Does this fit within pyDARN's scope? 
-  [x] Is this a minor change i.e., new small utility function?
-  [ ] Is this a major change i.e., new plotting function?
-  [ ] Do you need help developing it? 
-  [ ] Have you created a pyDARN GitHub [project](https://github.com/SuperDARN/pydarn/projects) to show the tasks needed to be done? 

## User Interface 

```python
pydarn.coord.SLANT_RANGE
```

PR #135, opened 2021-03-23
[EHN] Coordinate Enumerator for future coordinate systems

mts299:
# Scope 

This feature is only an enum to replace the boolean in *rtp.py* for `slant` 

**issue:** #133 

## Approval

**Number of approvals:** 2 : 1 tester  1  reviewer

## Test

```bash
>>> import pydarn
resetting environment variable IGRF_COEFFS in python script
resetting environment variable AACGM_v2_DAT_PREFIX in python script
non-default coefficient files may be specified by running aacgmv2.wrapper.set_coeff_path before any other functions
>>> pydarn.Coord.SLANT_RANGE
<Coord.SLANT_RANGE: 2>
>>> pydarn.Coord.RANGE_GATE
<Coord.RANGE_GATE: 1>
```

```python
import pydarn
import pydarnio

fitacf_file = './data/20100305.1401.00.rkn.fitacf'
pydarn.RTP.plot_summary(fitacf_data, beam_num=15)
plt.show()
```

Figure_1

Should produce as  normal with no differences


carleyjmartin:
Do you have an example of code that plots with range gates, I can't seem to get it to work?

carleyjmartin:
I like the idea of enum for the coordinate systems and units because we're eventually going to be getting more than two. The range gates/slant range are units of distance that you can plots to on a y-axis, I'm not sure they would fall into the same category as geographic/geomagnetic/polar coordinates, if this is where this enum is going? Unless this enum will just be a choice for units and the coordinate systems will be in another one?

carleyjmartin (approved):

Devin-Huyghebaert (approved):
Tested.  Works as expected.

Issue #130, opened 2021-03-22
45 km and 15 km fields to SuperDARN Radars


mts299:
# New Feature

**Name**: extra fields in Radars

**module**:  *superdarn_radars.py* 

**package**: *utils/*

## Scope

This will add 45 and 15 km max gates in the SuperDARN Radars table that can be used for reference in future development vs. using hardware files that are not consistent in their values. 

This will not have other max range gates for other types of gate sizes 

## Description

This will improve the inconsistent number of gates in files that don't store the value and something the user can grab from the radar table without needing a data file. This will also help with FOV and the FOV calculation

### pyDARN Checklist 

-  [x] Does this fit within pyDARN's scope? 
-  [x] Is this a minor change i.e., new small utility function?
-  [ ] Is this a major change i.e., new plotting function?
-  [ ] Do you need help developing it? 
-  [ ] Have you created a pyDARN GitHub [project](https://github.com/SuperDARN/pydarn/projects) to show the tasks needed to be done? 

## User Interface 

``` python
pydarn.superDARNRadars.radars[66].45_maxgate
```

PR #136, opened 2021-03-24
Ehn/radars gate field

mts299:
# Scope 

I only did a 45 km will do 15 km later, got the value from recent fitacf files. 

**issue:** #130 

## Approval

**Number of approvals:** 1

## Test

``` python
pydarn.SuperDARNRadars.radars[1].range_gate_45
100
```


carleyjmartin (commented):
Minor changes and one question that's only here because something else popped up in the changes for some reason that's not really related.

I just want to check, these are like a fall back value in case there is no value given from a data file (just plotting plain FOV/ data is from a grid file), and when a fitacf file is read in, if it has more than the given value here, the code later will take that into account and plot the FOV with the number of gates given in the fitacf file, if plotting a fan plot? 

I also thought that these values were going to be taken from the fov files when plotting rather than the hdw files as then when plotting a fov, it will never give an error? Then we only need to think about keeping the fov files up to date, rather than the files and these hardcoded values.

Devin-Huyghebaert:
@carleyjmartin brings up some good points and questions for this code.  Are these values to be used as a default?  I do worry about hard coding in values, requiring a new release of pyDARN to change.

carleyjmartin (approved):
Where should be document that this is an interim fix whilst we work on rbpos?

mts299:
Um I was gonnna add it to the FOV docs with rbpos added in so this could be merged and then we can do another full docs review on things as we now need to comment slant range and other things

mts299:
@carleyjmartin is this good to be merged? Need to use it for rpos code  :p

Issue #139, opened 2021-03-29
Fan plots drop the rightmost beam


SEC-VT:
# BUG 

In plotting/fan.py, the _plot_fan_ method does not plot the last beam in a scan.

## Priority

- [ ] Minor - typo, incorrect naming, fails on specific uncommon situation
- [x] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

## Example of the bug 

This is an example of the output from the plotting/fan.py/Fan._plot_fov_ method, overlaid with a scatter plot of the *thetas* and *rs* output. Note the extra beam clockwise of the FOV.
![image](https://user-images.githubusercontent.com/33470230/112887831-9fcbcc80-90a1-11eb-86c1-76f521abd070.png)
The same beam is dropped in the _plot_fan_ routine which actually plots the vectors.

### Attempts

I'm working on a PR that adds cartopy functionality to the fan plots which addresses this issue.

### Data Location

Using INV and BKS fitacfs from Virginia Tech.

## Potential Bug Location 

The issue stems from lines 299-314 (in _plot_fov_):
```python
        if boundary:
            # left boundary line
            plt.polar(thetas[0:ranges[1], 0], rs[0:ranges[1], 0],
                      color='black', linewidth=0.5)
            # top radar arc
            plt.polar(thetas[ranges[1] - 1, 0:thetas.shape[1] - 1],
                      rs[ranges[1] - 1, 0:thetas.shape[1] - 1],
                      color='black', linewidth=0.5)
            # right boundary line
            plt.polar(thetas[0:ranges[1], thetas.shape[1] - 2],
                      rs[0:ranges[1], thetas.shape[1] - 2],
                      color='black', linewidth=0.5)
            # bottom arc
            plt.polar(thetas[0, 0:thetas.shape[1] - 2],
                      rs[0, 0:thetas.shape[1] - 2], color='black',
                      linewidth=0.5)
```
The _right_ boundary line and _bottom_ arc are subtracting **2** from the shape, instead of **1**. The _top_ radar arc should not subtract anything.
Similarly, in line 183 (_plot_fan_ method):
```python
        # Begin plotting by iterating over ranges and beams
        for gates in range(ranges[0], ranges[1]-1):
            for beams in range(thetas.shape[1] - 2):
                # Index colour table correctly
                cmapindex = (scan[gates, beams] + abs(zmin)) /\
```
[...]

PR #140, opened 2021-03-29
Fix for dropped beam in plot_fan and plot_fov

billetd:
# Scope 

Fixes issue where a beam worth of data was dropped, and where the top arc of the radar FOV was plotted one range gate too low. Described in more detail in issue #139. 

## Approval

**Number of approvals:** 1

## Test

```python
file='20170731.2202.00.inv.fitacf'
fanplot = pydarn.Fan.plot_fan(fitacf_data)
plt.show()	
```
Compared with current develop branch, one more beam of data should be visible on the rightmost portion of the FOV

### DIFF
--- a/pydarn/plotting/fan.py
+++ b/pydarn/plotting/fan.py
@@ -179,8 +179,8 @@ class Fan():
             zmax = defaultzminmax[parameter][1]
 
         # Begin plotting by iterating over ranges and beams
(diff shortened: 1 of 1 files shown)


mts299:
Tested and looks good. @SEC-VT if you want to test and do the honors of pressing merge as new pyDARN dev go ahead :)

mts299 (approved):

PR #152, opened 2021-04-28
added costants file

mts299:
# Scope 

This file makes a  constants fille so  we share  the same constants across pyDARN 

## Approval

**Number of approvals:** 1

## Test

More of a review may be test slant range. 
``` python
import pydarn
import pydarnio

fitacf_file = './data/20210205.1200.00.cly.a.fitacf'
pydarn.RTP.plot_range_time(fitacf_data, beam_num=7)
plt.show()
```

Figure_1

### DIFF
--- a/pydarn/utils/constants.py
+++ b/pydarn/utils/constants.py
@@ -13,10 +13,10 @@
 # Modifications:
 #
 
-# Earth's radius (km)
+# Standard accepted IUGG global average of Earth's radius (km)
 Re = 6371
+# Earth radius at the equator (km)
 EARTH_EQUATORIAL_RADIUS = 6378.137
 
 # Speed of light (m/s)
 C = 2.99792458e+8
-
--- a/pydarn/utils/constants.py
+++ b/pydarn/utils/constants.py
@@ -1,0 +1,22 @@
+# (C) Copyright SuperDARN Canada, University of Saskatchewan
+# Author: Marina Schmidt
+#
+# Disclaimer:
+# pyDARN is under the LGPL v3 license found in the root directory LICENSE.md
+# Everyone is permitted to copy and distribute verbatim copies of this license
+# document, but changing it is not allowed.
+#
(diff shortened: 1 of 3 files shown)


pydarn/utils/constants.py
```diff
@@ -0,0 +1,22 @@
...
+# This version of the GNU Lesser General Public License incorporates the terms
+# and conditions of version 3 of the GNU General Public License,
+# supplemented by the additional permissions listed below.
+#
+# Modifications:
+#
+
+# Earth's radius (km)
+Re = 6371
+EARTH_EQUATORIAL_RADIUS = 6378.137
```
carleyjmartin, line 18:
```suggestion
# Standard accepted IUGG global average of Earth's radius (km)
Re = 6371
# Earth radius at the equator (km)
EARTH_EQUATORIAL_RADIUS = 6378.137
```
I'm not fussy, but I know some people might be, so added more description to the two Earth radii to explain the difference.
And just in case anyone asks: International Union of Geodesy and Geophysics (IUGG) suggests the average global value used as Re should be 6371km.  (Ironically, the International Astronomical Union (IAU) suggests using the equatorial radius as Re, however, we're not measuring distances in Earth radii in pyDARN, so I'm on the IUGG side for this.)

I guess if we ever need it we can add the polar radius when applicable - couldn't remember if we used the polar radius as a semi-minor axis in rpos code.

carleyjmartin (approved):
Just added some extra info in comments, imo it's fairly straight forward to understand before, but thought it might help if anyone has different preferences in what Re means to them later down the line. 

EDIT: Also no issues in testing!

mts299:
We used the equatorial one for the flattening but otherwise didn't use any other version.

carleyjmartin:
> We used the equatorial one for the flattening but otherwise didn't use any other version.

I thought so, but was too lazy to go find out myself 📦 

Merge time?

PR #146, opened 2021-04-01
Ehn/cartopy

SEC-VT:
# Scope 

This PR includes the following updates/features:
(What has been done in other PRs):
- [ ] Added Cartopy based plots
- [x] Added a check to determine if cartopy is installed correctly on a used machine
- [x] Added exception for when users without cartopy try to plot a cartopy plot
- [x] Reduced the overhead incurred from repeated calls to `plt.fill` with fewer calls to `plt.pcolormesh` in `plot_fan`
- [x] Added check for datetime input to `plot_fan`'s _scan_index_ parameter

Resolves #143 resolves #124 

## Approval

**Number of approvals:** 2

## Test

Here's what I'm using to test the different cases:

```python
import bz2
import pydarnio
import pydarn
import datetime as dt
import matplotlib.pyplot as plt
import matplotlib

matplotlib.use('tkagg')


def test():
    invfile = '20160306.1200.00.inv.fitacf.bz2'

    with bz2.open(invfile) as fitfile:
        fstream = fitfile.read()
        sdarn_read = pydarnio.SDarnRead(fstream, True)
        inv_records = sdarn_read.read_fitacf()


    kwargs = {'boundary': True,
              'zmax': 200,
              'zmin': -200,
              'groundscatter': True,
              'lowlat': 40,
              'scan_index': 34}

    # First, a regular polar plot
    pydarn.Fan.plot_fan(inv_records, colorbar=True, **kwargs)
    plt.show()
    # Next, a cartopy plot
    kwargs['polar'] =  False
    pydarn.Fan.plot_fan(inv_records, colorbar=True, **kwargs)
    plt.show()
    # datetime object handling!
    kwargs['scan_index'] =  dt.datetime(2016, 3, 6, 12, 33)
    pydarn.Fan.plot_fan(inv_records, colorbar=True, **kwargs)
    plt.show()
    # let's break something by plotting out of range
    try:
        kwargs['scan_index'] =  dt.datetime(2016, 3, 6, 18, 30)
        pydarn.Fan.plot_fan(inv_records, colorbar=True, **kwargs)
        plt.show()
    except Exception as err:
        print(err)
    # plotting to an axes already created
    fig = plt.figure('TEST Figure')
    ax = fig.add_subplot(111)
    kwargs['scan_index'] =  dt.datetime(2016, 3, 6, 12, 33)
    pydarn.Fan.plot_fan(inv_records, ax=ax, colorbar=True, **kwargs)
    plt.show()
    # plot FOV still works, right?
[...]

Issue #124, opened 2021-03-19
DateTime Object Input for Fan Plots


carleyjmartin:
# New Feature

**Name**: *plot_fan* 

**module**:  *fan.py* 

**package**: *plotting/*

## Scope

- [ ] Create new optional input for `plot_fan`  
- [ ] Conversion of datetime to `scan_index` (don't think it's generalised enough to go in utils, or would be used anywhere else, so probably a few lines of code in `plot_fan`?)
- [ ] Return the `scan_index` value for user to note/if it's needed elsewhere - possible need to be in utils?
- [ ] Update documentation on fan plotting.

*What will the feature do?*
- [ ] Allow plotting fan plots from a given date and time, rather than requiring a scan number.
- [ ] Convert a datetime object into the relevant scan index for given data file to be used in the plotting.

*What will the feature not do?* 
- [ ] Changing any code relating to scan_index and the underlying way that fan plots are plotted.
- [ ] Scan index will still be a parameter that you can input into fan plots.

## Description

New optional input for fan plots where a datetime object can be inputted which is used to find the scan index number at that time to plot the scan at the given time.
Do not intend to change how `plot_fan` works (I.e using `scan_index` to read in correct data) unless required.
Make as similar to `plot_grid` datetime input functionality as possible.
Be aware `scan_index` default is 1 - handle properly if optional input of datetime is given. 
Be aware of previous conventions used in other libraries (I.e RST gets the *next* scan from the time given) and write documentation accordingly.

Discussion subject: function of converting datetime to scan index required in a utils package, or just as small addition to `plot_fan`?

### pyDARN Checklist 

-  [x] Does this fit within pyDARN's scope? 
-  [x] Is this a minor change i.e., new small utility function?
-  [ ] Is this a major change i.e., new plotting function?
-  [ ] Do you need help developing it? 
-  [x] Have you created a pyDARN GitHub [project](https://github.com/SuperDARN/pydarn/projects) to show the tasks needed to be done?  (Added to fan plots project)

## User Interface 

```python
    if datetime given as input
        convert datetime object into scan_index in file
        return scan_index
   
    input new scan_index into remaining plot_fan code
```

PR #155, opened 2021-05-04
Ehn/datetime scanindex

mts299:
# Scope 

This PR will focus on the scan index using `datetime` object 

I got this modification from @SEC-VT PR #146 

**issue:** #124 

## Approval

**Number of approvals:** 1 this is a simple fix and test 
## Test

``` python                                                                                                                                                                                                   
import pydarn
from datetime import datetime

pydarn.Fan.plot_fan(fitacf_data, scan_index=datetime(2021, 2, 5, 12, 5)) 

plt.show()
```

Figure_1

### DIFF
--- a/pydarn/plotting/fan.py
+++ b/pydarn/plotting/fan.py
@@ -70,8 +70,9 @@ class Fan():
                 Default: Generates a polar projection for the user
                 with MLT/latitude labels
 
-            scan_index: int
+            scan_index: int or datetime
                 Scan number from beginning of first record in file
+                or datetime given first record to match the index
                 Default: 1
             parameter: str
                 Key name indicating which parameter to plot.
--- a/pydarn/plotting/fan.py
+++ b/pydarn/plotting/fan.py
@@ -22,12 +22,13 @@ import matplotlib.pyplot as plt
 import numpy as np
 
 from matplotlib import ticker, cm, colors
-from typing import List
+from typing import List, Union
 
 # Third party libraries
 import aacgmv2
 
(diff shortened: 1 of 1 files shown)


carleyjmartin (approved):
Works beautifully, noticed that if you give it a datetime with a non-zero second value you'll get the error below - I guess we don't need seconds, but probably good to document to only put in hours and minutes when writing the docs up for this?
```
File "/Users/carley/Documents/testing_code/venv-datatimefan/lib/python3.7/site-packages/pydarn/plotting/fan.py", line 151, in plot_fan
    raise plot_exceptions.IncorrectDateError(scan_datetimes[0],
NameError: name 'plot_exceptions' is not defined
```

mts299:
fixed the error not know `plot_exceptions` so now you should get the correct exception. 

@carleyjmartin should it get to the closes record if it exists?

carleyjmartin:
> @carleyjmartin should it get to the closes record if it exists?

Just writing up our discussion on this:
I think in 90% of cases the radars are running 1 min scans, if there is no scan for that minute it just returns the incorrect date error. Which is fine, I assume that it's better to say 'no data', than bump you to the nearest time without you realising it's not the exact time you put in. 

IMO this function is mainly only useful for normal scan anyway, other modes like a camping beam would looks pretty meaningless as a full scan. Like you said earlier, a solution for that would be being able to plot individual beams. 

I'm happy to merge with the note that we should just explain in the docs that fan plots only represent normal scan or similar data effectively, and you might need to guess a few times at the exact times of scan if it's not running an exactly 1 min scan mode - specialist scans can use a future iteration of plot_beam! :D

PR #156, opened 2021-05-05
added title functionality

mts299:
# Scope 

This PR creates titles for fan plots 

**issue:**

## Approval

**Number of approvals:** 1 super easy

## Test

``` python 
import pydarn

pydarn.Fan.plot_fan(fitacf_data, scan_index=10)


plt.show()
```

Figure_1

### DIFF
--- a/pydarn/plotting/fan.py
+++ b/pydarn/plotting/fan.py
@@ -55,7 +55,7 @@ class Fan():
                  groundscatter: bool = False,
                  zmin: int = None, zmax: int = None,
                  colorbar: bool = True,
-                 colorbar_label: str = '', title: str = ''):
+                 colorbar_label: str = '', title: bool = True):
         """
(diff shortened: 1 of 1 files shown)


carleyjmartin (commented):
Works well as default to plot, and works when passing own title string. 

Does the default title need an end time for fan plots? In the plots I've made it just give the same time as the start time/ if it was a one minute scan should it technically be a minute later? I don't know if it's worth trying to figure out what the time period of the scan is, maybe just leave it with the start time only?

mts299:
Yea I wasn't sure, I could add seconds?

carleyjmartin:
> Yea I wasn't sure, I could add seconds?

Seconds would be good, definitely be accurate then :P

carleyjmartin (approved):

PR #157, opened 2021-05-05
Ehn/fan speedup

mts299:
# Scope 

Gathers hemisphere information using SuperDARNRadars class. This is to set the standard so consistency for other PRs will follow a similar way of checking as we are getting a lot of hemisphere options passed in. 

## Approval 

**Number of approvals:** 1

## Test

``` python
import pydarn
from datetime import datetime

#fitacf_file = 'data/20210205.1200.00.mcm.a.fitacf'
#pydarn.Fan.plot_fov(66, datetime(2021, 2, 5, 12, 5), ranges=[0, 75])
#pydarn.Fan.plot_fov(4, datetime(2021, 2, 5, 12, 5), ranges=[0, 75])

pydarn.Fan.plot_fan(fitacf_data, scan_index=datetime(2021, 2, 5, 12, 5)) 

plt.show()
```

Stems from fov but I also tested with fan :p

### DIFF
--- a/pydarn/plotting/fan.py
+++ b/pydarn/plotting/fan.py
@@ -185,6 +185,8 @@ class Fan():
             zmin = defaultzminmax[parameter][0]
         if zmax is None:
             zmax = defaultzminmax[parameter][1]
+        norm = colors.Normalize
+        norm = norm(zmin, zmax)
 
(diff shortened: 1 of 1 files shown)


mts299:
incorrect branch

PR #159, opened 2021-05-06
EHN: Hemisphere  determination in Fan plots 

mts299:
# Scope 

This uses `SuperDARNRadars` to determine the hemisphere of the radar. 

## Approval

**Number of approvals:** 1 - changed 1 line

## Test

```  python
import pydarn
from datetime import datetime

pydarn.Fan.plot_fov(66, datetime(2021, 2, 5, 12, 5), ranges=[0, 75])
#pydarn.Fan.plot_fov(4, datetime(2021, 2, 5, 12, 5), ranges=[0, 75])

plt.show()
```
North
Figure_2

South
Figure_1

### DIFF
--- a/pydarn/plotting/fan.py
+++ b/pydarn/plotting/fan.py
@@ -28,7 +28,8 @@ from typing import List, Union
 import aacgmv2
 
 from pydarn import (PyDARNColormaps, build_scan, radar_fov, citing_warning,
-                    time2datetime, plot_exceptions)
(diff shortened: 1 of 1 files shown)


carleyjmartin (approved):
Tested using INV and MCM, no issues.

carleyjmartin:
I'm just getting completely blank fov plots, and I don't know why.

Issue #160, opened 2021-05-06
Polar Plotting in matplotlib v3.4.1


carleyjmartin:
# BUG 

Polar plotting functionality has changed with the new release of matplotlib version 3.4.1
If using the current plot_fov method - there will be no errors or warnings, it will just not plot. 
It does not effect plotting of data in fan plots. 

## Priority

- [ ] Minor - typo, incorrect naming, fails on specific uncommon situation
- [x] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

## Example of the bug 

It's just a blank polar plot when using plot_fov. 

### Attempts

No attempts made yet: I will attempt to look into it but we will have to consider how to make it work for older versions or force it to install 3.4.1 at install of pyDARN?

## Potential Bug Location 

fan.py - lines 322 - 337

PR #153, opened 2021-04-29
Polar plots modularity

billetd:
# Scope 

Remove the functionality of generating a default polar plot from each individual function which does it, placing it in it's own file called `axis.py`. This houses the polar plotting code `axis_polar`, accepting all normal options as kwargs from functions that call it.

Currently, only `plot_fan` and `plot_fov` benefits from this, but it will soon apply to `plot_grid` and any future "mapping" functions, like convection maps. `axis.py` will be further developed in the future so other types of projections, not just `axis_polar`, will be available. 

Currently, the only keywords which `axis_polar` accepts is `lowlat` and `hem` for the lower latitude boundary of the polar plot and the hemisphere, respectively. If no hemisphere is given, it works it out based on what radar is being plotted. 

## A few other changes
- `dtime` keywords have been standardised as `date`, as they were inconsistent.
- `plot_fov` now defaults to current time if no date is given. It would originally give an error.
- General cleaning up of typos and leftover comments which were no longer correct

## Approval

**Number of approvals:** 2

## Test

Both `plot_fan` and `plot_fov` take in the exact same keywords as they did before, except now they are handled as kwargs:
```python
# Read in data
file = "../pydarntesting/20200923.1001.00.rkn.fitacf"
SDarn_read = pydarn.SuperDARNRead(file)
fitacf_data = SDarn_read.read_fitacf()

# Plot the FOV of the radar with a lower latitude boundary of 60 degrees AACGM latitude 
beam_corners_aacgm_lats, beam_corners_aacgm_lons, thetas, rs, ax = \
		pydarn.Fan.plot_fov(stid=fitacf_data[0]['stid'], lowlat=60)
```
```python
# Make a similar style fan plot
beam_corners_aacgm_lats, beam_corners_aacgm_lons, scan, grndsct = \
	pydarn.Fan.plot_fan(fitacf_data, lowlat=50)
```
```python
# Generate a polar axis object for the southern hemisphere
ax = pydarn.Projections.axis_polar(lowlat=50, hem='S')
```

PR #163, opened 2021-05-11
switched polar to plot methods

mts299:
# Scope 

This addresses the bug with matplotlib 4.3.1 without it affect older versions

**issue:** #160 

## Approval

**Number of approvals:** 1 - @carleyjmartin  

## Test

plot fov
``` python
import pydarn
from datetime import datetime

pydarn.Fan.plot_fov(66, datetime(2021, 2, 5, 12, 5), ranges=[0, 75])

plt.show()
```

Figure_1

### DIFF
--- a/pydarn/plotting/fan.py
+++ b/pydarn/plotting/fan.py
@@ -321,20 +321,20 @@ class Fan():
 
         if boundary:
             # left boundary line
-            plt.polar(thetas[0:ranges[1], 0], rs[0:ranges[1], 0],
-                      color='black', linewidth=0.5)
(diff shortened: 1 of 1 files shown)


carleyjmartin (approved):
Tested using versions 3.2.1 and 3.4.1 of matplotlib for plot_fan and plot_fov. 

No issues, both versions and methods plotted as expected. 

I think the key to this working all round is line 309 `ax = plt.axes(polar=True)` so will need to check that this is included in PR #153 (which I think it is anyway, but it doesn't hurt to double check)

PR #169, opened 2021-06-03
Ehn/add kwargs

mts299:
# Scope 

This PR focuses on switching parameters to `kwargs` when it shares parameters 

**issue:**

## Approval

**Number of approvals:** 2
## Test

Should test all types of plotting methods 
- [ ] range-time 
- [ ] time-series
- [ ] summary plots
- [ ] grid 
- [ ] fan 
- [ ] fov


carleyjmartin (commented):
Only got around to testing a few things, will finish up on Monday.

carleyjmartin (commented):
Tested: 
[x] range-time - No issues, but careful when merging with the new changes in the release branch with the `coords`/`coord` keyword. And as an afterthought, I'll check on the release branch that the docs have the correct version. 
[x] time-series - No issues
[x] summary plots - see previous comment on error
[x] grid - No issues with single radar grid files - same issue as release branch with multiple radar grid files.
[x] fan - No issues
[x] fov - No issues

mts299:
Lets old off on this until after the patch release branch is made since this does change some keywords I think...

pydarn/plotting/grid.py
```diff
@@ -155,20 +137,20 @@ def plot_grid(cls, dmap_data: List[dict], record: int = 0,
...
             if time_diff.seconds/60 > time_delta:
                 raise plot_exceptions.NoDataFoundError(parameter,
                                                        start_time=start_time)
         else:
             record = 0
-            dtime = dt.datetime(dmap_data[record]['start.year'],
+            date = dt.datetime(dmap_data[record]['start.year'],
                                 dmap_data[record]['start.month'],
                                 dmap_data[record]['start.day'],
                                 dmap_data[record]['start.hour'],
```
carleyjmartin, line 156: It's out of scope of the changes so can't add a suggestion but there's a stray dtime on line 163

pydarn/plotting/grid.py
```diff
@@ -82,12 +81,6 @@ def plot_grid(cls, dmap_data: List[dict], record: int = 0,
             parameter: str
                 Key name indicating which parameter to plot.
                 Default: vel (Velocity). Alternatives: 'pwr', 'wdt'
-            lowlat: int
-                Lower AACGM latitude boundary for the polar plot
-                Default: 50
-            fov: bool
-                Set to false to not plot the outline of the FOV
-                Default: True
```
carleyjmartin, line 90: If I fix the dtime error I then get:
```
line 164, in plot_grid
    boundary=fov,
NameError: name 'fov' is not defined
```
I think maybe fov kw shouldn't be contained in kwargs as we want to be able to set a default if not given?

pydarn/plotting/grid.py
carleyjmartin: And on line 176

carleyjmartin: Also in line 165 - 170, these options should be changed to **kwargs instead. Except the boundary = fov as there is a name change going on there and if we're setting it to a default it should be set, see an above comment for that.

carleyjmartin (commented):
No issues with range-time, summary, time-series, fan, fov plots
There are a couple of stray `dtimes` in `grid.py`, and one call to `Fan.plot_fov` that needs `**kwargs` and checking on how we want to deal with `boundary=fov` if `fov` is in `kwargs` then it will break as we want it to give a default value, so should it be a keyword?

mts299:
I think I got it all resolved now @carleyjmartin if you want to retest

pydarn/plotting/grid.py
```diff
@@ -178,11 +160,9 @@ def plot_grid(cls, dmap_data: List[dict], record: int = 0,
             warnings.simplefilter("ignore")
             for stid in dmap_data[record]['stid']:
                 _, aacgm_lons, _, _, ax =\
-                        Fan.plot_fov(stid, dtime,
-                                     boundary=fov,
+                        Fan.plot_fov(stid, date,
                                      lowlat=lowlat,
                                      line_color=line_color,
```
carleyjmartin, line 165:
```suggestion
                        Fan.plot_fov(stid, date,
```
Sorry, I didn't explain it very well. lowlat, line_color, radar_location, and radar_label are all in kwargs, so if they're named here and you don't give them as a kw in the call to plot it will break. Just need to remove them and add **kwargs

pydarn/plotting/grid.py
```diff
@@ -178,11 +160,9 @@ def plot_grid(cls, dmap_data: List[dict], record: int = 0,
...
-                                     fov_color=fov_color,
                                      radar_location=radar_location,
                                      radar_label=radar_label,
                                      ax=ax)
```
carleyjmartin, line 168:
```suggestion
                                     ax=ax, **kwargs)
```
Sorry it made me split it up around the deleted line to be able to do it as a suggestion.
ax=ax still needs to be here as if it's a combined grid file it will make a new ax for each radar instead of putting them all together.

carleyjmartin (approved):

Issue #170, opened 2021-06-16
BUG: Groundscatter 'greying out' is effected by the lower limit of colour bar.


carleyjmartin:
# BUG 

If `groundscatter = True` is given in a range-time or summary plot then the amount of data that is greyed out is changed by the value if zmin (lowest value of colourbar) as below this value is grey. Examples below.
The good news is I think it's failing safe, i.e. no data is being ungreyed, just extra data is being greyed out so no one should be seeing groundscatter and interpreting it as echoes, but extra actual echoes are greyed out.

## Priority

- [ ] Minor - typo, incorrect naming, fails on specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation 
- [x] Urgent - failure in usage, large misrepresentation

## Example of the bug 

All examples showing data from:
```python
file = "/Users/carley/Documents/data/20131031.2201.00.sas.fitacf"
SDarn_read = pydarn.SuperDARNRead(file)
fitacf_data = SDarn_read.read_fitacf()
```
Ex1:  Ground scatter on, no limits set
`a = pydarn.RTP.plot_range_time(fitacf_data, beam_num=5,groundscatter=True)`
<img width="1029" alt="Screen Shot 2021-06-16 at 12 47 45 PM" src="https://user-images.githubusercontent.com/60905856/122275562-127a6600-cea1-11eb-8f9f-faef3a95d181.png">

Ex2: Ground scatter off, no limits set
`a = pydarn.RTP.plot_range_time(fitacf_data, beam_num=5,groundscatter=False)`
<img width="1022" alt="Screen Shot 2021-06-16 at 12 48 56 PM" src="https://user-images.githubusercontent.com/60905856/122275701-3d64ba00-cea1-11eb-9f34-76f479b31e5b.png">

Ex3: Ground scatter on, limits set high
`a = pydarn.RTP.plot_range_time(fitacf_data, beam_num=5,groundscatter=True, zmin=-2000, zmax=2000)`
<img width="1027" alt="Screen Shot 2021-06-16 at 12 50 33 PM" src="https://user-images.githubusercontent.com/60905856/122275911-769d2a00-cea1-11eb-85d5-b9366fbf10be.png">

Ex4: Ground scatter on, limit set low
`a = pydarn.RTP.plot_range_time(fitacf_data, beam_num=5,groundscatter=True, zmin=-200, zmax=200)`
[...]

PR #172, opened 2021-06-18
FIX: ground scatter bug

mts299:
# Scope 

This bug fixes the over-the-coloring of ground scatter, sadly could not get grey into the plot as a legend but could look at that later. 

**issue:** #170 

## Approval

**Number of approvals:** 1 simple fix

## Test

**matplotlib version**: 3.3.4

```
import pydarn

fitacf_file = './data/20150308.1400.03.cly.fitacf'

pydarn.RTP.plot_range_time(fitacf_data, groundscatter=True, background='black', zmin=-200, zmax=200, beam_num=3, parameter='v')

plt.show()
```

`develop`

develop_plot

`fix/groundscatter`

test_plot

### DIFF
--- a/pydarn/plotting/rtp.py
+++ b/pydarn/plotting/rtp.py
@@ -5,7 +5,7 @@
 #
 # Modifications:
 # 2021-05-12 Francis Tholley added gate2grounscatter to range-time plots
-#
+# 2021-06-18 Marina Schmidt (SuperDARN Canada) fixed ground scatter colour bug
 # Disclaimer:
 # pyDARN is under the LGPL v3 license found in the root directory LICENSE.md
 # Everyone is permitted to copy and distribute verbatim copies of this license
--- a/pydarn/plotting/rtp.py
+++ b/pydarn/plotting/rtp.py
@@ -412,7 +412,6 @@ class RTP():
             im2 = ax.pcolormesh(time_axis, y_axis, ground_scatter, lw=0.01,
                            cmap=gs_color, norm=norm, **kwargs)
 
-            #cmap.set_under(groundscatter)
         elif groundscatter:
             ground_scatter = np.ma.masked_where( z_data != -1000000, z_data)
             gs_color = colors.ListedColormap(['grey'])
--- a/pydarn/plotting/rtp.py
+++ b/pydarn/plotting/rtp.py
@@ -398,10 +398,6 @@ class RTP():
                      'elv': PyDARNColormaps.PYDARN}
             cmap = cmaps[parameter]
 
-        if isinstance(groundscatter, str):
-            cmap.set_under(groundscatter, 1.0)
-        elif groundscatter:
-            cmap.set_under('grey', 1.0)
 
         # set the background color, this needs to happen to avoid
         # the overlapping problem that occurs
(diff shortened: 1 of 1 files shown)


pydarn/plotting/rtp.py
```diff
@@ -398,17 +398,27 @@ def plot_range_time(cls, dmap_data: List[dict], parameter: str = 'v',
...
         cmap.set_bad(color=background, alpha=1.)
         # plot!
         im = ax.pcolormesh(time_axis, y_axis, z_data, lw=0.01,
                            cmap=cmap, norm=norm, **kwargs)
+
+        if isinstance(groundscatter, str):
+            ground_scatter = np.ma.masked_where( z_data != -1000000, z_data)
+            gs_color = colors.ListedColormap([groundscatter])
+            im2 = ax.pcolormesh(time_axis, y_axis, ground_scatter, lw=0.01,
+                           cmap=gs_color, norm=norm, **kwargs)
```
carleyjmartin, line 413: Just want to check this is working as expected. If groundscatter gets given a string instead of a bool character, then it plots ground scatter in grey. From line 411, I would expect that if the string was say 'black' it would plot it in black, but it doesn't, it plots in grey still? Is the grey the expected outcome, and this is kinda a way to catch any user input errors?

And as a side, if you use the string 'False' by accident instead of the bool it plots the gs in grey.

pydarn/plotting/rtp.py
```diff
@@ -398,17 +398,27 @@ def plot_range_time(cls, dmap_data: List[dict], parameter: str = 'v',
...
+
+            #cmap.set_under(groundscatter)
```
carleyjmartin, line 415:
```suggestion
```

carleyjmartin (approved):
Works great, doesn't plot any extra data as grey, and also solves the other issue I was having with it keeping the old colour map when re-plotting in the same script and switching gs on and off! I had one question, which isn't an issue, just a question I also had with the old way of doing it too. 

matplotlib: 3.4.2
python: 3.7.3
Mac OS Catalina 10.15.5

Tested using: 
```python
import pydarn
import datetime as dt

file = "/Users/carley/Documents/data/20131031.2201.00.sas.fitacf"

a = pydarn.RTP.plot_summary(fitacf_data, beam_num=5, groundscatter=True)
plt.show()
a = pydarn.RTP.plot_summary(fitacf_data, beam_num=5, groundscatter=False)
plt.show()
```

first plot: 
Screen Shot 2021-06-18 at 1 45 02 PM


second plot:
Screen Shot 2021-06-18 at 1 45 12 PM

pydarn/plotting/rtp.py
mts299: So I tried plotting in black and 'False' both give me expected behaviour. 

What is the code you are using?

carleyjmartin: Is the expected behaviour that it should plot out the groundscatter as black then, if `groundscatter='black'`?

Just the code above but with 
`a = pydarn.RTP.plot_summary(fitacf_data, beam_num=5, groundscatter='black')`
and
`a = pydarn.RTP.plot_summary(fitacf_data, beam_num=5, groundscatter='False')`

Both give me GS in grey as if it's testing the truthiness of the two (same as first plot above).
I'm gonna go out on a limb and say it's possibly a matplotlib thing again, so will downgrade matplotlib and test again too.

carleyjmartin: Never mind, it's mysteriously working as expected now!

Issue #179, opened 2021-07-06
[EHN] pytest


mts299:
# New Feature

**Name**: *TestModule*

**module**:  *test_modulename.py* 

**package**: *test*

## Scope

Currently making a pytest suite for pyDARN to test options don't crash the code. 
It will focus mostly on passing setting options in and returned values. 

Will not check if the plot is correct, this will need to be done by users.  

## Description

This feature will be a test suite of unit tests for various plotting methods to ensure when changes are made the parameters of the method are not affected the same as the results. utils will also be tested. 

### pyDARN Checklist 

-  [x] Does this fit within pyDARN's scope? 
-  [x] Is this a minor change i.e., new small utility function?

PR #184, opened 2021-08-06
Pytest/rtp

mts299:
# Scope 

Adding RTP pytest tests, focuses on just testing the RTP plotting methods and parameters passed in with the test data. However, I still need to add some documentation and imagine the folder structure may need to change but not sure yet. 

I also added some fixes already!

**issue:** #179 

## Approval

**Number of approvals:** 2 - review and tester

## Test

**matplotlib version**: 3.3.4
**Note testers: please indicate what version of matplotlib you are using**

*provide a list or code to test*

```
cd pydarn/test
pytest test_RTP.py 
```

PR #186, opened 2021-08-11
Pytest/structure

mts299:
# Scope 

This PR focuses on the ability to call it from the main pydarn directory which is why travis does

**issue:** #179 

## Approval

**Number of approvals:** 1 - simple change

## Test

**matplotlib version**: 3.3.4

```
cd pydarn
pytest --ignore=test/data
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,4 +50,4 @@ install:
     - pip3 install .
     - pip3 install pytest
     - pip3 install pydarnio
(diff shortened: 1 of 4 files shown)


carleyjmartin (approved):
For solely running from the `pydarn` directory rather than in `test`, this works perfectly. 
No issues outside of ones relating to #185 and #184.

PR #192, opened 2021-09-08
Fix develop pytest errors

mts299:
# Scope 

After merging pytest into the patch release there were some fails in the `pytest` so fixed them. 
The major issue is when `rpos` code needed single range gates to be converted to slant ranges so I added an option for both but didn't get changed in RTP.  Hopefully this fixes your other issues @carleyjmartin 

**issue:**

## Approval

**Number of approvals:** *1*

## Test

**matplotlib version**:  3.3.4
You may want to do some plotting with RTP to make sure it looks correct but also running `pytest`

````
cd pydarn
pytest
````
Should not fail


carleyjmartin (approved):
👌 

pytest runs smoothly, no issues with any prewritten code used to test before

PR #194, opened 2021-09-09
EHN: Channel keyword for Fan plots

### LINKED ISSUE #182: BUG: Fan plot issues for files with multiple channel data
# BUG 

Issue plotting fan plots and rtp using data from PYK on 2015-03-08 noted by @Shirling-VT in last two releases.

Issue is rooted in how we grab a scan, in pyDARN we look for 1 in the scan flag to denote that a new scan has started, but on this day, the scan flag has a pattern (1,1,0,0,0,0....0,1,1,0,0...) where we see that it's only going to plot the first beam before thinking it's going to the next on the next beam. Resulting in only plotting one beam:
image

The scan mode is 153, and other 153 data seems to work fine, is this an issue with this specific radar/datafile/mode?

There is also an issue with beam 5 on this day for the same data, where it appears all data is uniform:
image

Again, is this an issue in the data or pyDARN specifically for this beam and day.

This time period appears to be fine for both issues on the VT website plotting tools that use DaViT, but we don't know if the data files are identical that are being used.

It would be good to test using our data files for DaViT (I believe @billetd is testing this) and I believe Pasha Ponomarenko (not in this repo?) is trying it out on his own software. Would be great to run it through RST too to see any differences - difference in plotting is that RST looks to plot 60 seconds of data and doesn't use the scan falg value like we do?

## Priority

- [x] Minor - typo, incorrect naming, fails on specific uncommon situation - appears to be specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation 

## Information
**python version**: NA
**OS**: Mac OS Catalina 
**matplotlib version**: 3.3.4 and 3.4.2

## Potential Solution(s)

If it turns out this is a data issue, then it's out of our hands, but a proposed function to specify to plot only one beam and the user can build up a fan plot of any beams they like (also useful for camping beam ect).

I do not have any input on the beam 5 issue as it appears that might be a difference in data in the files?

If anyone has these issues on other data it would be great to find some more to test on!

Further discussion of this issue can be found in release PRs #181 and #166

carleyjmartin:
# Scope 

This PR adds a keyword to the plot_fan method which allows the user to plot data from a single channel.
In addition, an exception has been made for when the user chooses a channel but it doesn't exist. 

A warning has also been made which warns that the data chosen has partial records: this is mainly due to the HAN and PYK radars running channels 1 and 2, but channel 2 only has partial records (CPID: 26999 'dummy channel'). But, the warning itself may be useful elsewhere down the line. 

The default will remain at `channel ='all'` which is the same as the RTP plots, and for the majority of cases 'all' will be fine and nothing will change- especially for the channels which are separated into files anyway. 

The main issue is that if you have files with multiple channel data, it may still try to plot 'all' and get confused and plot partial or no data (like the example in the issue), the only thing I can think of to remove this is to set a warning to say 'there's more than one channel of data in here, take care with interpretation and use the channel keyword if needed'. However, this will need to search all the data read in to see which channels are used, and can considerably slow down fan plotting where this isn't really needed 95% of the time. If you think this is a good idea, I can add in this warning, otherwise the user will just have to figure out that the data has more than one channel in the file and that's why it's looking a bit funky. 

**issue:** close #182 

## Approval

**Number of approvals:** 2

## Test

**matplotlib version**: 3.4.2
This will return a plot of the channel 1 data:
```python
import pydarn
import datetime as dt

file = "/Users/carley/Documents/data/20141218.0801.00.han.fitacf"

time = dt.datetime(2014,12,18,8,24)

fanplot = pydarn.Fan.plot_fan(fitacf_data, scan_index=time,
                    colorbar_label='Velocity [m/s]', lowlat=50, groundscatter=True, channel=1)            
plt.show()
```
Screen Shot 2021-09-09 at 3 56 35 PM

This will return an empty plot and should show the warning that it's a partial record:
```python
import pydarn
import datetime as dt

file = "/Users/carley/Documents/data/20141218.0801.00.han.fitacf"

time = dt.datetime(2014,12,18,8,24)

fanplot = pydarn.Fan.plot_fan(fitacf_data, scan_index=time,
                    colorbar_label='Velocity [m/s]', lowlat=50, groundscatter=True, channel=0)            
plt.show()
```
Screen Shot 2021-09-09 at 3 57 06 PM

Screen Shot 2021-09-09 at 3 57 24 PM

And this will give the exception that the channel doesn't exist:
```python
import pydarn
import datetime as dt

file = "/Users/carley/Documents/data/20141218.0801.00.han.fitacf"

time = dt.datetime(2014,12,18,8,24)

fanplot = pydarn.Fan.plot_fan(fitacf_data, scan_index=time,
                    colorbar_label='Velocity [m/s]', lowlat=50, groundscatter=True, channel=0)            
plt.show()
```
Screen Shot 2021-09-09 at 4 05 37 PM

This will return the original plot that gets confused (channel='all'):
```python
import pydarn
import datetime as dt

file = "/Users/carley/Documents/data/20141218.0801.00.han.fitacf"

time = dt.datetime(2014,12,18,8,24)

fanplot = pydarn.Fan.plot_fan(fitacf_data, scan_index=time,
                    colorbar_label='Velocity [m/s]', lowlat=50, groundscatter=True)            
plt.show()
```
Screen Shot 2021-09-09 at 3 58 27 PM

Note! that most radars use 0 as the main channel, but a lot use 1 and 2 if using two channels, so you never know.
The HAN data is just an example, but I've tested on SAS a and b files along with single channel files, please test with whatever you have available to get a rounded picture of it working. 
Pytest ran with no failures, not modified at this time.

### DIFF
--- a/pydarn/exceptions/plot_exceptions.py
+++ b/pydarn/exceptions/plot_exceptions.py
@@ -186,9 +186,12 @@ class NoChannelError(Exception):
     This error is raised when the channel specified by user
     does not exist
     """
-    def __init__(self, channel: int):
+    def __init__(self, channel: int, opt_channel: int = None):
         self.channel = channel
-        self.message = "There is no data for channel {channel}"\
-                       .format(channel=self.channel)
+        self.opt_channel = opt_channel
+        self.message = "There is no data for channel {channel},"\
+                       " try channel {opt_channel}."\
+                       .format(channel=self.channel,
+                       opt_channel=self.opt_channel)
         super().__init__(self.message)
         pydarn_log.error(self.message)
--- a/pydarn/exceptions/plot_exceptions.py
+++ b/pydarn/exceptions/plot_exceptions.py
@@ -1,5 +1,9 @@
 # Copyright (C) SuperDARN Canada, University of Saskatchewan
 # Authors: Marina Schmidt
+#
+# Modifications:
+# 20210909: CJM - Added NoChannelError
+#
 # Disclaimer:
 # pyDARN is under the LGPL v3 license found in the root directory LICENSE.md
 # Everyone is permitted to copy and distribute verbatim copies of this license
@@ -173,5 +177,18 @@ class OutOfRangeGateError(Exception):
             " between 0 - {max_gate}".format(gate_num=self.gate_num,
                                              param=self.parameter,
                                              max_gate=self.max_range_gate-1)
+        super().__init__(self.message)
+        pydarn_log.error(self.message)
+
+
+class NoChannelError(Exception):
+    """
+    This error is raised when the channel specified by user
+    does not exist
+    """
+    def __init__(self, channel: int):
+        self.channel = channel
+        self.message = "There is no data for channel {channel}"\
+                       .format(channel=self.channel)
         super().__init__(self.message)
         pydarn_log.error(self.message)
--- a/docs/user/fan.md
+++ b/docs/user/fan.md
@@ -1,6 +1,7 @@
 <!--Copyright (C) SuperDARN Canada, University of Saskatchewan 
 Author(s): Daniel Billet 
 Modifications:
+20210922: CJM - included info on new channel option
 
 Disclaimer:
 pyDARN is under the LGPL v3 license found in the root directory LICENSE.md 
@@ -89,7 +90,8 @@ Here is a list of all the current options than can be used with `plot_fan`
 
 | Option                        | Action                                                                                                  |
 | ----------------------------- | ------------------------------------------------------------------------------------------------------- |
-| scan_index=(int or  datetime) | Scan number or datetime, from start of records in file                                                  |
+| scan_index=(int or  datetime) | Scan number or datetime, from start of records in file corresponding to channel if given                |
+| channel=(int or 'all')        | Specify channel number or choose 'all' (default = 'all')                                                |
 | groundscatter=(bool)          | True or false to showing ground scatter as grey                                                         |
 | ranges=(list)                 | Two element list giving the lower and upper ranges to plot, grabs ranges from hardware file (default [] |
 | cmap=matplotlib.cm            | A matplotlib color map object. Will override the pyDARN defaults for chosen parameter                   |
@@ -105,6 +107,11 @@ Here is a list of all the current options than can be used with `plot_fan`
 | radar_label=(bool)            | Places the radar 3-letter abbreviation next to the radar location                                       |
 | kwargs **                     | Axis Polar settings. See [polar axis](axis.md)                                                          |
 
+
+!!! Note
+    For some control programs, the user may need to specify a channel integer as `'all'` will not correctly show the data.
+    In other cases, the user may want to specify the channel and use an integer (N) for the `scan_index`. Be aware that this will show the
+    data for the Nth scan of only the chosen channel, not that of the entire file. 
 
 `plot_fan` can concatenate with other plots including itself, here is an example of plotting two different radars with some of the above parameters:
 
--- a/docs/user/range_time.md
+++ b/docs/user/range_time.md
@@ -92,6 +92,7 @@ To see all the customisation options, check out all the parameters listed in 'rt
 |------------------------------|-------------------------------------------------------------|
 | start_time=(datetime object) | Control the start time of the plot                          |
 | end_time=(datetime object)   | Control the end time of the plot                            |
+| channel=(int or string)      | Choose which channel to plot. Default is 'all'.             |
 | groundscatter=(bool)         | True or false to showing ground scatter as grey             |
 | date_fmt=(string)            | How the x-tick labels look. Default is ('%y/%m/%d\n %H:%M') |
 | zmin=(int)                   | Minimum data value to be plotted                            |
--- a/pydarn/__init__.py
+++ b/pydarn/__init__.py
@@ -23,6 +23,7 @@ from .exceptions import radar_exceptions
 from .exceptions.warning_formatting import standard_warning_format
 from .exceptions.warning_formatting import only_message_warning_format
 from .exceptions.warning_formatting import citing_warning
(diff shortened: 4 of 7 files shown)


mts299:
@carleyjmartin Code looks good, can you add a pytest for this parameter change?

pydarn/exceptions/plot_exceptions.py
```diff
@@ -175,3 +179,16 @@ def __init__(self, parameter: str, gate_num: int, max_range_gate: int):
...
+
+class NoChannelError(Exception):
+    """
+    This error is raised when the channel specified by user
+    does not exist
+    """
+    def __init__(self, channel: int):
+        self.channel = channel
+        self.message = "There is no data for channel {channel}"\
+                       .format(channel=self.channel)
```
mts299, line 192: Similar to my no data warning maybe we could suggest a channel that data has. This would help with say "twofsound" when we have 1 and 2 and when we have stero mode with 0 and 1

mts299:
So I got some twofsound data and the channels  change per scan but when I do a `scan_index` and switch between 1 and 2 it comes up with data and not one with "no data" error
``` python
import pydarn
from datetime import datetime

rkn_file = 'data/20180101.0000.01.rkn.fitacf'
rkn_data = pydarn.SuperDARNRead().read_dmap(rkn_file)
ydarn.Fan.plot_fan(rkn_data, scan_index=3,
                    colorbar=False, line_color='blue',
                    radar_label=True, channel=1)
```

`channel=1`
channel1

`channel=2`
channel2

mts299 (changes requested):
Look into the twofsound scan index issue

pydarn/exceptions/plot_exceptions.py
carleyjmartin: I've added the option to give an option in the exception, but the implementation of the exception in fan.py is a bit rudimentary at the moment, assuming I will probably change how it works when I sort out the other issue below. 


(I also know why it's doing that other issue, but need to have a think about how to go around it so it preserves the fix for the issue with the HAN data too)

carleyjmartin:
So here's the dilemma with this: at the moment if a channel is given in the function call for a fan, then the code will remove any data that isn't in that channel and then proceed with the rest of the fan plotting. This works well for HAN and PYK, as their scans are interleaved with each other i.e:
scan flag:1,1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0...
channel:  0,1,0,1,0,1,0,1,0,1,0,1,0,1,0,1,0,1,0,1,0,1,0,1,0,1,0,1,0,1,0,1...
So the channel switches between each data point in the file, kind of blurring what it means to have a '1st' scan, where is it the full cycle above that is one scan or two scans? As you can see, just removing one channel is fine for this issue as we're not too set on numbering of the scans as they start/end at roughly the same time, and selecting channel 1 or 0 will give you a nice 16 beams scan. 

HOWEVER, we then have the opposite issue with RKN twofsound data, where the scan and channels look like this:
scan flag: 1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,1,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0...
channel:   1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2,2...
So using channel='all' doesn't break here, and we can just use that, but then if you select only channel 1 and rather than inputting a time, input a scan index N, it will give you the Nth scan of ONLY that channel data, not the entire data set. But inputting a time will either plot it correctly or tell you there's no data as it's the other channel operating at that minute, which is as expected.
The Nth scan index issue can be fixed by just including a channel check later in the if statement (same as RTP) and removing the current solution. But, then we would be back to it not working for the HAN and PYK channel/scan-flag pattern. I'm not sure if it's possible to include both of these solutions without some pretty horrific if loops and/or convoluted pattern detection that will considerably slow the plotting down - we can't use cpid detection as the cpid can be different for each channel.

I might suggest that we can specify that the option to 'choose a channel' is for scans that are operating on two channels simultaneously rather than one scan after the other, as the one after the other scans can easily be accessed using the channel='all' option, whereas the simultaneous scans can't and require some filtering of data to plot. I'll keep having a poke at it, and if there are any suggestions let me know but as of now I can't think of a non-convoluted way of keeping both of these solutions in the code without further confusion or issues.

Anyway, I hope that makes sense. TLDR: combination of different scan_flag and channel patterns mean there's no single all encompassing solution to plotting a specific channel/scan for fan plots.

mts299:
@carleyjmartin still reading the novel response, but shouldn't we look for the scan first then plot if the channel is correct? Would this not fix the problem. 

I worry if we remove the channel data then find the scan then we are not looking accurately at scan 3 anymore

carleyjmartin:
@mts299  The main issue I can't think around is how to pick out the scans from the sequentially scanning data? Is the 'first scan' the entire '1,1,0,0,0,0,0,0,0,0,0,0,..' or is that technically two scans? How can we tell the code to sometimes ignore that second 1? Which I guess is the original issue of scan_flag being used to build the scans, when in this case the scan_flag is only signalling the new scan for a single channel, and then the next record goes back to a previous scan. 

I think I have an idea anyway, will have another crack tomorrow.

mts299:
Okay message me on slack for more idea thinking.

carleyjmartin:
Added channel option to fan plot read the docs, also added a note under the options to discuss that in some cases you might NEED to use a channel integer, not the default 'all'. And that the scan_index will be counting scans of the specific channel if a channel is given. This is changed in the docs and in the docstring in fan.py. Slightly out of scope, but very minor change: also added the channel option the range-time docs, I think there was confusion long ago about RTP taking a channel option, and it was just accidentally missed off in the docs.

mts299 (approved):

PR #196, opened 2021-09-14
Median Filter

shibaji7:
# Scope 

Reduce salt and paper noise in the fitacf type dataset.

## Approval

**Number of approvals:** 2 (test and code review)


mts299:
@shibaji7 I think you set up the PR incorrectly by merging develop into boxcar instead of boxcar into develop. 

Also, I couldn't find the boxcar code, where would you have put it?

Issue #205, opened 2021-11-01
Adding library attributes


mts299:
# New Feature

**Name**: `__version__` `__bibtext__` and more

**module**:  *__version.py* and others

## Scope

Looking to add some library attributes like `pydarn.__version__` realizing we don't have this when I needed to double-check the version on cedar. 

## Description

This will require a new file(s), I am also looking at adding the version file but there are other attributes I will look at adding. Please list any that you are interested in adding. 

### pyDARN Checklist 

-  [x] Does this fit within pyDARN's scope? 
-  [x] Is this a minor change i.e., a new small utility function?

PR #213, opened 2021-11-19
[EHN] adding version info

mts299:
# Scope 

Added `version.py` and `__version__` ability to pydarn to allow users to know which version is installed on their systems. 

**issue:** #205 

## Approval

**Number of approvals:** 1 

## Test

```
import pydarn
pydarn.__version__
```
returns 2.2.1 

Also when you install this branch it should refer to the version as 2.2.1

### DIFF
--- a/pydarn/__init__.py
+++ b/pydarn/__init__.py
@@ -13,6 +13,9 @@ module, classes, and functions.
 
 import os
 
+# version file
+from .version import __version__
+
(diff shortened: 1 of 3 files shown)


carleyjmartin (approved):
No issue on pip3 install

Tested as
```
import pydarn
pydarn.__version__
```
And returns 
`2.2.1`

Issue #227, opened 2022-02-25
AACGM doesn't work for fan plots


mts299:
# BUG 

`Coords.AACGM` currently is not working as it only plots `Coords.GEOGRAPHIC` but works for `coords.AACGM_MLT`

Found this when working on adding `Coords.GROUND_SCATTERED_MAPPED` to fan plots. 

## Priority

- [ ] Minor - typo, incorrect naming, fails on specific uncommon situation
- [x] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

## Information
**python version**: 3.6
**OS**: ubuntu
**matplotlib version**: 3.5.1

## Example of the bug 

``` python
pydarn.Fan.plot_fan(data, grid=True, scan_index=5, fov_color='blue', coords=pydarn.Coords.AACGM_MLT, radar_label=True, groundscatter=True)
plt.show()
```
![aacgm](https://user-images.githubusercontent.com/6520530/155799738-d8fc7881-e0e8-485c-9233-ccb61de78456.png)

For `Coords.GEOGRAPHIC`

![geo](https://user-images.githubusercontent.com/6520530/155799808-ea8c7bb7-d1ba-4b47-b071-e77b73040c9b.png)

For `Coords.AACGM_MLT`

![aacgm_mlt](https://user-images.githubusercontent.com/6520530/155799818-cbdbf343-30d9-4013-9f1d-5dc66da89a50.png)

### Potential issue

https://github.com/SuperDARN/pydarn/blob/87fb29db71f4af37f9670ac66b6ba07beeb4abe4/pydarn/utils/radar_pos.py#L104

Needs to include `Coords.AACGM`

PR #228, opened 2022-02-25
fixed the missing coords.AACGM

mts299:
# Scope 

This is to fix the issue of using `Coords.AACGM` with fan plots
**issue:** #227 

## Approval

**Number of approvals:**  1 
## Test

**matplotlib version**: 3.5.1
 ``` python
pydarn.Fan.plot_fan(data, grid=True, scan_index=5, fov_color='blue', coords=pydarn.Coords.AACGM, radar_label=True, groundscatter=True)
plt.show()
```

aacgm


mts299:
Correction this doesn't fix it entirely because the gridding is wrong 

I am working on cleaning up radar_pos with a refactoring with coordinates so go to close this branch as the other one will do a better cleaner job

PR #232, opened 2022-03-02
FIX: plot_fan variable return order changes and boundary option redo

### LINKED ISSUE #231: BUG: Boundary option removed in plot_fov
# BUG 

This is a really really tiny bug, but I thought before I forget about it it needs to go back in so that we can have the option to not plot the fov in methods that call plot_fov.

Basically, the updates to include cartopy just forgot that `boundary` option exists so it's not used in plot_fov anymore in develop branch. 

(While I'm at it I'll also fix the minor bug caused by plot_fov changing the return values discussed after merging #202 )

## Priority

- [x] Minor - typo, incorrect naming, fails on specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation 

## Potential Bug Location 

You can see what it used to be on main branch in fan.py line 376.

## Potential Solution(s)

Just add it back in and hope it doesn't break cartopy options?

carleyjmartin:
# Scope 

This PR aims to solve the bugs caused by two branches being merged into develop at similar times. 
The first bug is the accidental removal of the `boundary` key word use in plot_fov/
The second bug is the changes to the variable return order for plot_fov which breaks the new maps.py plotting.
I've also removed the second citation warning that we don't need now it shows up on import. 

If anyone finds some other bugs due to this 'double merge' let me know and I can fix it in this PR.

**issue:** to close #231 

## Approval

**Number of approvals:** 2 please

## Test

**matplotlib version**: 3.5.1

both can be tested using maps, boundary True or False, the fact it plots at all means the second fix works:
```python
import datetime as dt
import pydarn

pydarn.Maps.plot_mapdata(map_data,record=180, 
            lowlat=60,
            boundary=False)

plt.show()
```
Screen Shot 2022-03-02 at 9 52 48 AM

If you want to test for fans, please try lots of different kwargs! :) :
```python
import pydarn
file = '/Users/carley/Documents/data/20211008.2000.00.rkn.a.fitacf'

#pydarn.Fan.plot_fan(fitacf_data, grid=True, scan_index=2, fov_color='blue',
#                    radar_label=True, groundscatter=True, boundary=False)
                    
pydarn.Fan.plot_fan(fitacf_data, boundary=False, scan_index=2, fov_color='blue',
                    radar_label=True, groundscatter=True,coords=pydarn.Coords.GEOGRAPHIC ,projs=pydarn.Projs.GEO_COASTALINE)
    
plt.show()
```

### DIFF
--- a/pydarn/plotting/fan.py
+++ b/pydarn/plotting/fan.py
@@ -331,7 +331,7 @@ class Fan():
 
     @classmethod
     def plot_fov(cls, stid: str, date: dt.datetime,
-                 ax=None, ccrs=None, ranges: List = [],
+                 ax=None, ccrs=None, ranges: List = [], boundary: bool = True,
                  rsep: int = 45, frang: int = 180,
                  projs: object = Projs.POLAR,
                  coords: Coords = Coords.AACGM_MLT,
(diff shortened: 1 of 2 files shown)


billetd:
For map related plots, I think everything to do with the individual sites (e.g. dots where sites are) should be off by default. When convection contours are eventually fully implemented, the default should probably just be contours and fit vectors.

mts299 (approved):

Issue #229, opened 2022-03-01
Refactor: radar_pos and utils folder


mts299:
# REFACTORING

**Feature:** Refactoring the following modules and maybe more to allow more flexibility, readability, and usability. 

**Module:**  *radar_pos.py* *coordinates.py* *conversions.py* (**new**) *range_estimations.py* 
 
**package** */utils*

## Reason

While working on #203 I found a lot hard to read code flow, coupling, mistakes, and bloated parameter lists. So this will handle some of this with encapsulation and abstraction. 

## Replacement

Removed SLANT_RANGE, GSMR, and RANGE_GATE to be placed in *range_estimations.py* under *Range_Estimation* class

## Details

This will only affect RTI plots and *radar_fov* however the calling of that function should not change.

PR #242, opened 2022-03-14
Ehn/kwargs radar pos

mts299:
# Scope 

This PR focuses on changing the module (file) structure and names of class in the utils folder to make a for better developer usage, organization, and for preventing bugs by less `if` checks on various coordinate systems. 

I had to modify `Fan` and `RTP` classes plotting methods for the new `range_estimations` parameter. 

**issue:** #229 #227 replaces PR #203

# What Changed?

- [x] **New** `range_estimation.py` contains only the range estimation methods like `gate2slant` and `gate2ground_scatter_mapped` 
- [x] **New** `RangeEstimation` for `enums` of `SLANT_RANGE`, `RANGE_GATE`, and `GSMR`
- [x] **Deprecated** `radar_fov`
- [x] **New** coordinate methods for each coordinate system: `aacgm_coordinates`, `aacgm_mlt_coordinates`, and `geographic_coordinates`
- [x] **New** *geo.py* store the geographical calcuations that were originally in *radar_pos.py*
- [x] **Change** `Projs.GEO_COASTALINE` -> `Projs.GEO` 
- [x] **New** GSMR works with fan plots - possibly grid and map as well?
- [x] **Update** now various functions take the required parameters and pass kwargs through 
- [x]  **Docs** should be updated with the above changes for the user's point of view 
- [x] **New** All enums so `RangeEstimation.SLANT_RANGE` actually maps to a function call so you can treat like a function in development and in use :)  

I think this is everything? 

## Approval

**Number of approvals:** 2 - 1 tester 1 reviewer 

## Test

**matplotlib version**:  3.5.1

## To Test

- [ ] `pytest`
- [ ] `plot_range_paramter`
- [ ] `plot_time_series`
- [ ] `plot_summary`
- [ ] `plot_fov`
- [ ] `plot_fan`
- [ ] Review documentation 

## Testing pytest

Please see documentation on how to run pytest. Note you will need a git clone version of the code on the correct branch to do this. 

## `RTP` Testing 

```python
import bz2 
import pydarn
import datetime as dt

cly_file = "/home/marina/superdarn/data/fitacf/20220108.0000.00.cly.a.fitacf.bz2"
sas_file = "/home/marina/superdarn/data/fitacf/20220108.0000.00.sas.a.fitacf.bz2"
with bz2.open(cly_file) as fp: 
    fitacf_stream = fp.read()

cly_data = pydarn.SuperDARNRead(fitacf_stream, True).read_fitacf()

with bz2.open(sas_file) as fp: 
    sas_stream = fp.read()

sas_data = pydarn.SuperDARNRead(sas_stream, True).read_fitacf()

pydarn.RTP.plot_range_time(cly_data, beam_num=7, range_estimation=pydarn.RangeEstimation.GSMR)
plt.show()

pydarn.RTP.plot_summary(cly_data, beam_num=7, groundscatter=True)
plt.show()
```
Figure_1

summary

## `Fan` Testing

```python
import bz2 
import pydarn
import datetime as dt

data_file = "/home/marina/superdarn/data/fitacf/20211008.2000.00.rkn.a.fitacf.bz2"
with bz2.open(data_file) as fp: 
    fitacf_stream = fp.read()

data = pydarn.SuperDARNRead(fitacf_stream, True).read_fitacf()

pydarn.Fan.plot_fan(data, grid=False, scan_index=5, radar_label=True, range_estimation=pydarn.RangeEstimation.GSMR, groundscatter=True)
plt.show()

pydarn.Fan.plot_fan(data, grid=False, scan_index=5, ranges=[70,90], radar_label=True, range_estimation=pydarn.RangeEstimation.GSMR,               groundscatter=True)
plt.show()

pydarn.Fan.plot_fan(data, grid=False, scan_index=5, radar_label=True, groundscatter=True, range_estimation=pydarn.RangeEstimation.GSMR,           coords=pydarn.Coords.GEOGRAPHIC, projs=pydarn.Projs.GEO, colorbar_label="Velocity m/s")
plt.show()
```
fan_1
fan_2
fan_3

**I will be refactoring grid and map modules in other PRs to keep this one focused**


docs/user/coordinates.md
```diff
@@ -15,13 +15,15 @@ the additional permissions listed below.
 # Coordinate Systems 
 ---
 
-pyDARN offers several different coordinate systems and units for plotting: 
+pyDARN offers several different measurement systems for various type of plotting:
```
carleyjmartin, line 18:
```suggestion
pyDARN offers several different measurement systems for various types of plotting: 
```

docs/user/coordinates.md
```diff
@@ -15,13 +15,15 @@ the additional permissions listed below.
 # Coordinate Systems 
 ---
 
-pyDARN offers several different coordinate systems and units for plotting: 
+pyDARN offers several different measurement systems for various type of plotting: 
+- Range Estimation: the estimate of how far the target (echo) to the radar
```
carleyjmartin, line 19:
```suggestion
- Range Estimation: the estimate of how far the target (echo) is from the radar
```

docs/user/coordinates.md
```diff
@@ -15,13 +15,15 @@ the additional permissions listed below.
...
+- Coordinates systems: Generic geographic coordinate systems with conversions to AACGM and AACGM MLT
```
carleyjmartin, line 20:
```suggestion
- Coordinate systems: Generic geographic coordinate system with conversions to AACGM and AACGM MLT 
```

docs/user/coordinates.md
```diff
@@ -46,6 +48,10 @@ Geographic plots include: fan, grid and convection map (coming soon) plots
 
 **AACGM**: `Coords.AACGM` is [Altitude Adjusted Corrected Geogmagnetic Coordinates developed by Dr. Simon Shepherd](http://superdarn.thayer.dartmouth.edu/aacgm.html) are an extension of corrected geomagnetic (CGM) coordinates that more accurately represent the actual magnetic field. In AACGM coordinates points along a given magnetic field line are given the same coordinates and thus are a better reflection of magnetic conjugacy. pyDARN uses AACGM-V2 from the [aacgmv2 python library](https://pypi.org/project/aacgmv2/). Implemented by Dr. Daniel Billett from University of Saskatchewan. 
 
-**Ground Scatter Mapped Geographic** (Coming Soon)
+**AACGM_MLT**: `Coords.AACGM_MLT` is `Coords.AACGM` with the longitude shift to line up
```
carleyjmartin, line 51:
```suggestion
**AACGM_MLT**: `Coords.AACGM_MLT` is `Coords.AACGM` with the geomagnetic longitude converted to magnetic local time
```

docs/user/coordinates.md
```diff
@@ -46,6 +48,10 @@ Geographic plots include: fan, grid and convection map (coming soon) plots
 
 **AACGM**: `Coords.AACGM` is [Altitude Adjusted Corrected Geogmagnetic Coordinates developed by Dr. Simon Shepherd](http://superdarn.thayer.dartmouth.edu/aacgm.html) are an extension of corrected geomagnetic (CGM) coordinates that more accurately represent the actual magnetic field. In AACGM coordinates points along a given magnetic field line are given the same coordinates and thus are a better reflection of magnetic conjugacy. pyDARN uses AACGM-V2 from the [aacgmv2 python library](https://pypi.org/project/aacgmv2/). Implemented by Dr. Daniel Billett from University of Saskatchewan. 
 
-**Ground Scatter Mapped Geographic** (Coming Soon)
+**AACGM_MLT**: `Coords.AACGM_MLT` is `Coords.AACGM` with the longitude shift to line up 
+
+`RangeEstimation` methods can be used with `Coords` calculation for example, using `Coords.GEOGRAPHIC` using `RangeEstimation.GSMR` to determine the coordinates.
```
carleyjmartin, line 53:
```suggestion
`RangeEstimation` methods can be used with a `Coords` calculation. For example, using `Coords.GEOGRAPHIC` and `RangeEstimation.GSMR` together, will give a plot of ionospheric echoes at a distance from the radar calculated in ground scatter mapped range, in geographic coordinates. 
```

carleyjmartin (commented):
This is a lot of work, and the restructuring is great! I found some pretty minor plotting things, and the non-geo axes plotting might be off a bit? Most (if not all) of the comments are suggestions for the docs.

MPL: version 3.5.1
Cartopy: version 0.20.2
python3: version 3.9.9

- [x] `pytest` passed all tests, one warning about np.str being depreciated
- [x]  `plot_range_time` all runs nicely without error, tested on three range estimation enums. Noticed that the velocity colour bar is not centered on zero, but when checking against develop branch it was also there so not a big issue.
- [x] `plot_time_series` -not tested alone, assuming any issue would show in summary plots
- [x] `plot_summary` no issues, tested on the three range estimation enums  
- [x] `plot_fov` - no issues, although I can't plot a second fan on the same axis if in GEO
- [x] `plot_fan` - runs without error, but noticed that the background of the data is now white instead of transparent - this can block the latitude lines and the continents:
Screen Shot 2022-03-15 at 10 49 05 AM
Screen Shot 2022-03-15 at 10 49 15 AM
This does not happen in FOV so assuming it's a thing in plot_fan:
Screen Shot 2022-03-15 at 10 49 29 AM

I was also comparing to develop branch, is the fan plot default now a different coordinate system? There is almost a 180 degree difference in the plots:
Screen Shot 2022-03-15 at 11 01 19 AM
left=develop, right=this branch


I haven't tested any maps or grids, and I'm guessing we might have to update the docs for RTP and Summary to add in the range_estimation kwarg.

KhanalKrishna:
I have tested the plots. Plots are produced as aspected. Comments are below each plots.

image

image

image

- Latitude longitude grid is not visible at the radar field of view. 

image

- Changing coordinate arugument changes the radar orientation and location but MLTs remain the same. Not sure how to use different coordinate systems.


image

- It would be useful to have ranges for beams too (like range-gates).

image


- It would be useful to have longitude/latitude grid according to the cooordinate system being used. I am not sure if there that capability already
- Changing the coord argument to AACGM gives the upper figure where radar orientation changes but the map remains the same. How to use a different coordinate system?

mts299:
@KhanalKrishna thank you for testing! 

- [x] Created a issue and will add the lat and lon lines in another PR to keep the scope of this one 
- [ ] FOV bug seems to not happen on my OS so I am gonna see if @carleyjmartin can look into it with a new PR as this happens on `develop`
- [ ] AACGM and cartopy is in the works... I will add a error message (again another PR) until we figure it out just so we can make a release sooner than later. 

I don't understand what you mean here:
> It would be useful to have ranges for beams too (like range-gates).

carleyjmartin:
@KhanalKrishna 
Could you let me know what you're working with? 
- matplotlib version
 - python version
 - OS
 - cartopy version
If possible, Thanks!

Shirling-VT:
MacOS 10.15.7
matplotlib 3.4.2
python 3.8.0

1. Pass pytest with a warning on pydarnio of np.str:
image

2. plot_range_paramter works fine with a warning on invalid value encountered in sqrt in range_estimations.py.
3. plot_time_series works fine.
4. plot_summary works fine.
5. plot_fov works fine.
6. plot_fan found the same issue of non-transparent fov:
image
image

carleyjmartin (approved):
So, I've looked into the range_estimation.py returning a warning for the use of np.sqrt
The error only occurs when you use GSMR units for plotting in RTP or Fan plots, and is caused because in the first few range gates the slant range is calculated to be less than the reflection height required for GSMR calculations. The function as such returns a NaN for these values as it is impossible to calculate a sqrt of a negative number. 
This is expected behaviour and the outcome is what we want later on, the NaN's are passed to the Fan or RTP plotting functions and are dealt with in there (they're basically removed and not plotted). We need the NaN's to keep the shape of the correct number of gates as it's passed to the plotting functions. It's just unfortunate that we now get a warning for that function, even though it's doing what we want it to do. This expected behaviour can be added to the documentation to reduce any bug reports on it, I'm not sure if there is a work around solution because we want those NaNs.


And just to address other concerns brought up by me and others:
- white fov is fixed in develop now after #247 was merged
- an error message has been set for the combination use of coords=aacgm and projs=geo so you can't plot it and it says why #246 
- lon/lat grid for geo is now made and in #245 
- the discrepancy in develop and this branch is fixed in this branch - develop is technically wrong
- Beam selection like range - out of scope here, but useful so an issue should be made for this.

Still to do: 
- @KhanalKrishna can you reply with the code used to create the two plots that resulted in the comment:  *Changing coordinate argument changes the radar orientation and location but MLTs remain the same. Not sure how to use different coordinate systems.* I don't get the same, so I think this can be solved by some good docs!

So all in all, either solved by other PR's or out of scope, so happy to approve for now.

KhanalKrishna:
Dear all,
Sorry for my late response. I was in hospital. Just got our newborn home.
@mts299,
"It would be useful to have ranges for beams too (like range-gates)". I meant if we could plot something like below where only 8 beams are plotted:
image

@carleyjmartin,
Here is the code I used:

import bz2 
import pydarn
import datetime as dt

cly_file = "20180204.1601.00.cly.fitacf.bz2"
sas_file = "20180204.1601.00.inv.fitacf.bz2"
with bz2.open(cly_file) as fp: 
    fitacf_stream = fp.read()

cly_data = pydarn.SuperDARNRead(fitacf_stream, True).read_fitacf()

with bz2.open(sas_file) as fp: 
    sas_stream = fp.read()

sas_data = pydarn.SuperDARNRead(sas_stream, True).read_fitacf()


pydarn.Fan.plot_fan(cly_data, grid=True, scan_index=5, radar_label=True, groundscatter=True, range_estimation=pydarn.RangeEstimation.GSMR, coords=pydarn.Coords.AACGM, projs=pydarn.Projs.GEO, colorbar_label="Velocity m/s")
plt.show()

pydarn.Fan.plot_fan(cly_data, grid=True, scan_index=5, radar_label=True, groundscatter=True, range_estimation=pydarn.RangeEstimation.GSMR, coords=pydarn.Coords.GEOGRAPHIC, projs=pydarn.Projs.GEO, colorbar_label="Velocity m/s")

plt.show()

carleyjmartin:
Thanks @KhanalKrishna, and absolutely ignore us on here, you have something more important to worry about! 
Congratulations :)

PR #258, opened 2022-04-28
EHN: Coastlines and Reference Vector Added to Convection Maps

carleyjmartin:
# Scope 

This PR includes the addition of an optional coastline map underlay using cartopy using a new keyword as it's in magnetic coordinates it's more involved than just adding an ax.coastlines() like in the geo fan plots. Because of the transformation, there is no available option to colour in the land/sea.
In developing this, I also added a keyword to allow the user to 'turn off' the coloring of the vectors as it looked like a huge mess with the coastlines, the vectors and the potential outlines, so giving the user more options to make it look nice past the default. At this moment, there is only the option to turn off the colormap and plot in a dark grey colour. We could add another key word which would allow us to color them, but there was already enough in this PR. 
In turning off the colour too, we needed a reference vector. So I've added the code for a reference vector, this vector is only plotted when the colormap isn't used automatically, no key word needed.

**issue:** NA

## Approval

**Number of approvals:** 2
I've included the cartopy logic for those who don't have cartopy, if possible please test with and without.

## Test

**matplotlib version**: 3.5.1

This set of keywords I would recommend for anyone wanting to plot the coastlines as it looks the nicest, but please test with various key word combos even if they look terrible! I have tested on Southern data: I _think_ it looks as it should, we're looking through the earth down to antarctica so I think it makes sense - please let me know!
The default is still the same default as before with no coastlines and grey potential contours.

```python
import pydarn
import datetime as dt

pydarn.Maps.plot_mapdata(map_data,record=0, 
            lowlat=50, coastline=True, contour_fill=True, color_vectors= False)
plt.show()
```
Screen Shot 2022-04-28 at 9 45 44 AM

```python
map_file = "/Users/carley/Documents/data/maps/20170113.s.map"
```
Screen Shot 2022-04-28 at 10 14 16 AM

### DIFF
--- a/pydarn/plotting/maps.py
+++ b/pydarn/plotting/maps.py
@@ -80,7 +80,7 @@ class Maps():
                      parameter: Enum = MapParams.FITTED_VELOCITY,
                      record: int = 0, start_time: dt.datetime = None,
                      time_delta: float = 1,  alpha: float = 1.0,
-                     len_factor: float = 150, color_vectors: bool = True, 
+                     len_factor: float = 150, color_vectors: bool = True,
                      cmap: str = None, colorbar: bool = True,
                      colorbar_label: str = '', title: str = '',
                      zmin: float = None, zmax: float = None,
@@ -214,7 +214,7 @@ class Maps():
             (aacgmv2.convert_mlt(aacgm_lons[0, 0], date) * 15)
         shifted_lons = data_lons - shifted_mlts
         # Note that this "mlons" is adjusted for MLT
-        mlons = np.radians(shifted_lons) 
+        mlons = np.radians(shifted_lons)
         mlats = data_lats
 
         # If the parameter is velocity then plot the LOS vectors
@@ -246,17 +246,18 @@ class Maps():
             v_mag = dmap_data[record]['vector.wdt.median']
             azm_v = np.radians(dmap_data[record]['vector.kvect'])
 
-        
-
         if parameter in [MapParams.FITTED_VELOCITY, MapParams.MODEL_VELOCITY,
                          MapParams.RAW_VELOCITY]:
             # Make reference vector and add it to the array to
             # be calculated too
-            reflat = plt.gca().get_ylim()[1] - 5
+            reflat = (np.abs(plt.gca().get_ylim()[1]) - 5) * hemisphere.value
             reflon = np.radians(45)
             v_mag = np.append(v_mag, 500)
-            azm_v = np.append(azm_v, np.radians(135))
-            
+            if hemisphere == Hemisphere.North:
+                azm_v = np.append(azm_v, np.radians(135))
+            else:
+                azm_v = np.append(azm_v, np.radians(45))
+
             # Angle to "rotate" each vector by to get into same
             # reference frame Controlled by longitude, or "mltitude"
             alpha = np.append(mlons, reflon)
@@ -266,7 +267,6 @@ class Maps():
             # Convert initial positions to Cartesian
             start_pos_x = (90 - abs(mlats)) * np.cos(mlons)
             start_pos_y = (90 - abs(mlats)) * np.sin(mlons)
-
 
             # Resolve LOS vector in x and y directions,
             # with respect to mag pole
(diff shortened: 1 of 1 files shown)


carleyjmartin:
Found a better way to do this, will commit code later today in another PR and update this one for just the reference vector and vector colours.

carleyjmartin:
This PR now only includes updates for the reference vector and removing the colour map to only plot in grey - as described above but without the coast lines - for coastlines go to #261. You can test with the below code:

Probably only requires one approval now as mainly aesthetic changes. 

MPL version used: 3.5.1

```python
import pydarn

#map_file = "/Users/carley/Documents/data/maps/20170113.s.map"
#map_file = "/Users/carley/Documents/data/maps/20190421.s.map2"

# Default default (radar fov turned off)
            lowlat=50, color_vectors=True)
plt.show()

            lowlat=50, color_vectors=False,
            contour_fill=True)
plt.show()
```
Which produces:
Screen Shot 2022-05-20 at 1 20 12 PM
Screen Shot 2022-05-20 at 1 20 43 PM

pydarn/plotting/maps.py
```diff
@@ -64,12 +66,14 @@ def plot_mapdata(cls, dmap_data: List[dict], ax=None,
...
-                     colorbar: bool = True, colorbar_label: str = '',
-                     title: str = '', zmin: float = None, zmax: float = None,
+                     len_factor: float = 150, color_vectors: bool = True,
+                     cmap: str = None, colorbar: bool = True,
+                     colorbar_label: str = '', title: str = '',
+                     zmin: float = None, zmax: float = None,
                      hmb: bool = True, boundary: bool = False,
                      radar_location: bool = False, map_info: bool = True,
-                     imf_dial: bool = True, **kwargs):
+                     imf_dial: bool = True, reference_vector: bool = False,
```
billetd, line 75: In my opinion the reference vector should be on by default, even if the vectors are coloured.

Additionally, it would be good if the magnitude of the reference vector itself was an option. Instead of `True` or `False`, `reference_vector: int = 500`. This way it will just always default to 500, and the user can input `False` if they don't want it.

pydarn/plotting/maps.py
```diff
@@ -134,6 +142,10 @@ def plot_mapdata(cls, dmap_data: List[dict], ax=None,
             imf_dial: bool
                 If True, draw an IMF dial of the magnetic field clock angle.
                 Default: True
+            reference_vector: bool
+                If True include reference vector and label in
+                lower right corner
+                Default: False
```
billetd, line 148:
```suggestion
                If True include reference velocity vector and 
                label in lower right corner
                Default: False
```

or if my above suggestion is used:

```suggestion
                Velocity magnitude to use for the
                reference vector, drawn in lower
                right corner
                Default: 500
```

pydarn/plotting/maps.py
carleyjmartin: Sounds good, I've made it so that the vector is on by default, and is controlled by entering an integer as `reference_vector = 1000` to set the size of the vector. It is on and 500 m/s as default, and can be turned off by setting the reference_vector to 0 or False.

The following code makes the three following plots (a:100m/s, b:off, c:default):
```python

# Default default (radar fov turned off)
            lowlat=50, color_vectors=True, reference_vector=1000)
plt.show()

            lowlat=50, color_vectors=False,
            contour_fill=True, reference_vector=False)
plt.show()

            lowlat=50, color_vectors=False,
            contour_fill=True)
plt.show()
```
Screen Shot 2022-06-17 at 2 58 21 PM
Screen Shot 2022-06-17 at 2 58 52 PM
Screen Shot 2022-06-17 at 2 59 20 PM

billetd (approved):

Shirling-VT:
@carleyjmartin Looks like none of the PRs I tried today works for me. Is it because I forgot how to do testing after not testing for a while? The reference vector does not show up, and changing it to other numbers (e.g. 1000) won't work (i.e., no reference vector at all). Seting color_vecotors to False won't work, the colorbar of vectors will always be there. Again, let me know if it's just me, I'll start over and reinstall pydarn to see if it's only issue on my end.

Issue #259, opened 2022-04-28
New Feature: Improvements to ACF Plotting


carleyjmartin:
# New Feature

**Name**: acf_power and acf_phase calculation and plotting

**module**:  in acf.py

**package**: plotting

## Scope

*What will the feature do?*
Plot the phase and power of the ACF plots.
We should also look into distinguishing between 'bad' lags and 'blanked' lags.

## Description
New methods to plot the phase and power of the ACF plots, recommended by scientists that these values will be helpful alongside the real and imaginary parts. The phase is given as atan(Im/Re) - this will need to be on an additional degrees/radians axes. The power is given as the sqrt(Im^2 + Re^2). 

### pyDARN Checklist 

-  [x] Does this fit within pyDARN's scope? 
-  [x] Is this a minor change i.e., new small utility function?
-  [ ] Is this a major change i.e., new plotting function?
-  [x] Do you need help developing it? 
-  [ ] Have you created a pyDARN GitHub [project](https://github.com/SuperDARN/pydarn/projects) to show the tasks needed to be done? 

### Development help

*pyDARN does not have a dedicated development team on hand to full feature development. 
However, if you are able to provide some time and help, the community will try to aid you the best way they can in making your feature possible.*

*If you need help please indicate what expertise you are looking for:*
- [ ] developer help with design and integration of code
- [x] scientific help with ensuring what you develop is scientifically correct/accurate
- [ ] project management to help organize what steps need to be taken to make this possible 

## User Interface 

Additional options in the acf plotting method.

## Extra Notes

If there is no interest in developing this, I can have a go at it, I don't think it should be something that takes up too much time.

PR #260, opened 2022-05-05
EHN: Power and Phase Added to ACF Plots

carleyjmartin:
# Scope 

This PR includes additions to the ACF plotting, as plots for phase and power of the ACF were requested along side. I've designed it similar to the Davit plots, but without the extra info that you get from fitacf.
The defaults have been changed to be un-normalized and for the power and phase to be plotted, but you can get the original by setting these to True and False respectively. 
I've also changed the logic a little so that the power plot does not get normalized. I've also added a return so the values of power and phase will be returned so the user can play around with them however they wish.

After discussions, we've found that the blanked lags are correct, so I've not done anything for that. We don't have the information to plot any extra lags (like the CRI bad lags) so we're leaving it at the basic Tx blanked lags. 

**issue:** #259 

## Approval

**Number of approvals:** 2 (please test for lots of variations on the plot_blank, normalization, and pwr_and_phs key words)

## Test

**matplotlib version**: 3.5.1

```python
import pydarn
from datetime import datetime

reader = pydarn.SuperDARNRead("/Users/carley/Documents/data/20140105.1208.03.rkn.rawacf")
data = reader.read_rawacf()
plt.figure(figsize=(12, 7))
pydarn.ACF.plot_acfs(data, beam_num=7, gate_num=44, normalized=False,
                start_time=datetime(2014, 1, 5, 12, 8), pwr_and_phs=True,
                plot_blank=True)
plt.show()
```
Produces:
Screen Shot 2022-05-05 at 1 49 13 PM

  - copyrighted (if applicable)

### DIFF
--- a/pydarn/plotting/acf.py
+++ b/pydarn/plotting/acf.py
@@ -220,48 +220,52 @@ class ACF():
                 raise plot_exceptions.OutOfRangeGateError(parameter, gate_num,
                                                           record['nrang'])
 
-        if normalized:
-            blank_re /= record['pwr0'][gate_num]
-            blank_im /= record['pwr0'][gate_num]
-
-            re /= record['pwr0'][gate_num]
-            im /= record['pwr0'][gate_num]
-
         # generates gaps where there are nan's
         masked_re = np.ma.array(re)
         masked_re = np.ma.masked_where(np.isnan(masked_re), masked_re)
 
         masked_im = np.ma.array(im)
         masked_im = np.ma.masked_where(np.isnan(masked_im), masked_im)
-        
-        if pwr_and_phs is True:
-            pwr = np.sqrt(np.square(masked_re) + np.square(masked_im))
-            phs = np.arctan2(masked_im, masked_re)
-            
-            masked_pwr = np.ma.array(pwr)
-            masked_pwr = np.ma.masked_where(np.isnan(masked_pwr), masked_pwr)
-            masked_phs = np.ma.array(phs)
-            masked_phs = np.ma.masked_where(np.isnan(masked_phs), masked_phs)
+
+        # Calculate pwr and phs regardless of choice as
+        # in return statement
+        pwr = np.sqrt(np.square(masked_re) + np.square(masked_im))
+        phs = np.arctan2(masked_im, masked_re)
+
+        masked_pwr = np.ma.array(pwr)
+        masked_pwr = np.ma.masked_where(np.isnan(masked_pwr), masked_pwr)
+        masked_phs = np.ma.array(phs)
+        masked_phs = np.ma.masked_where(np.isnan(masked_phs), masked_phs)
+
+        if normalized is True:
+            masked_re /= record['pwr0'][gate_num]
+            masked_im /= record['pwr0'][gate_num]
+
+            blank_re_n = blank_re / record['pwr0'][gate_num]
+            blank_im_n = blank_im / record['pwr0'][gate_num]
 
         # plot real and imaginary with power
         if pwr_and_phs is True:
             fig = plt.gcf()
-            gs = fig.add_gridspec(2,2)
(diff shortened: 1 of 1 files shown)


KhanalKrishna:
Matplotlib version: 3.5.1

#script
import pydarn
from datetime import datetime

reader = pydarn.SuperDARNRead("20140101.0001.00.rkn.rawacf")
data = reader.read_rawacf()
plt.figure(figsize=(12, 7))
pydarn.ACF.plot_acfs(data, beam_num=8, gate_num=44, normalized=False,
                start_time=datetime(2014, 1, 1, 12, 8), pwr_and_phs=True,
                plot_blank=True)
plt.show()


image

I tried for 4-5 random data files and different beams. Same error each time.

carleyjmartin:
Hey @KhanalKrishna you might have to change the minute and second values in `start_time=datetime(2014, 1, 1, 12, 8)` as it is looking for a time of 12:08 which doesn't exist in the file you're using.

Shirling-VT:
MacOS 10.15.7
matplotlib 3.4.2
python 3.8.0

import bz2 
import pydarn
from datetime import datetime

file = "/Users/xueling/data/superdarn/2014/rawacf/rkn/20140105.1208.03.rkn.rawacf.bz2"
rawacf_data =  pydarn.SuperDARNRead().read_dmap(file)

plt.figure(figsize=(10, 6))
pydarn.ACF.plot_acfs(rawacf_data, beam_num=7, gate_num=44, normalized=False,
                     start_time=datetime(2014, 1, 5, 12, 8), pwr_and_phs=True,
                     plot_blank=True)
plt.show()

Got the following error:
image

If not adding the pwe_and_phs keyword, I am able to get the upper panel:
image

I'm thinking about this might be a matlibplot version issue.

carleyjmartin:
Hi @Shirling-VT thanks for testing. I downgraded my matplotlib to 3.4.2 and it still works as expected. I can't recreate the issue that you have found without running the code on develop branch, which makes sense because the keyword doesn't exist on there, I don't want to assume you haven't but have you definitely tested the branch titled `ehn/acf_pwr`?

ecbland:
The new keywords are working for me. I'm using `matplotlib 3.3.4`. 

Here is a beautiful ACF from Longyearbyen...
image

I'm not sure why the ACF y-axis range is not set correctly. Notice that both the real and imaginary parts of the ACF for lag=1 are outside the axis range. This also happens for `normalized=False`. 

Is it possible to force the x-axis to have integer values? No problem if not. 

----------
> We don't have the information to plot any extra lags (like the CRI bad lags) so we're leaving it at the basic Tx blanked lags.

Just a note that "bad lags" due to CRI or ACF shape are identified based on subjective criteria that were removed in FitACF3.0. The only lags that are "bad" are the Tx-overlap lags. So this is as far as we should be going anyway :smile:

Shirling-VT:
@carleyjmartin I'm pretty sure I'm on the right branch, I also checked the codes locally and can see those changes consistent with the file change on this branch. I tested multiple times today and still got the same error on not recognizing pwr_and_phs as a keyword. It might be just me due to an odd combination of matplotlib and python. Feel free to merge it if it works fine for everyone else.

carleyjmartin:
@ecbland Thanks Emma, you found a bug! I had the calculation for the auto detected limits the wrong way around (it's now max(abs()) instead of abs(max())!) I've also added two lines to fix all the x-axes to only plot the integers. New plot below!
Thanks for clarifying the bad lags!

Screen Shot 2022-07-14 at 1 10 37 PM


@Shirling-VT I have no idea how to help make it work, I guess it could be a problem with so many pydarn version installed?  It seems to work for all the matplotlib versions allowed by the install that I've tried, so I have no clue why it's not picking the new keyword :/ I really don't think the python version difference would cause this.

ecbland (approved):

PR #266, opened 2022-07-03
comments out setting min start on beam_corners_lats and beam_corners_lons in geo_coordinates function

francistee26:
# Scope 
This PR comments out # y0inx = np.min(np.where(np.isfinite(beam_corners_lats[:,0]))[0]) in the geo_coordinates function in coordinates.py. Since we need to return all of the beam_corners_lats and beam_corners_lons.

## Approval

**Number of approvals:** 3

## Test
```
import pydarn
import pydarnio

from datetime import datetime
from pydarn import Coords

sps_fitacf_file = 'data_vt_sps/20170124.0000.01.sps.a.fitacf'

sps_fitacf_reader = pydarnio.SDarnRead(sps_fitacf_file)
sps_fitacf_data = sps_fitacf_reader.read_fitacf()

fig = plt.figure(figsize=(12,12))
kwargs1 = {
'boundary': True,
'zmax': 200,
'zmin': -200,
'groundscatter': False,
'scan_index':22,
'colorbar_label':'Power [dB]',
'line_color':'red',
'radar_label': True,
'parameter': 'p_l',
'coords': Coords.GEOGRAPHIC
}

ax, beam_corners_lats, beam_corners_lons, scan, grndsct = \
    pydarn.Fan.plot_fan(sps_fitacf_data, colorbar=True, **kwargs1)
    
plt.show()
```

### DIFF
--- a/pydarn/utils/coordinates.py
+++ b/pydarn/utils/coordinates.py
@@ -55,8 +55,8 @@ def geo_coordinates(stid: int, beams: int = None,
                                                 **kwargs)
             beam_corners_lats[gate-gates[0], beam] = lat
             beam_corners_lons[gate-gates[0], beam] = lon
-    y0inx = np.min(np.where(np.isfinite(beam_corners_lats[:,0]))[0])
-    return beam_corners_lats[y0inx:], beam_corners_lons[y0inx:]
+    # y0inx = np.min(np.where(np.isfinite(beam_corners_lats[:,0]))[0])
+    return beam_corners_lats, beam_corners_lons
 
 
 def aacgm_coordinates(stid: int, beams: int = None, gates: tuple = None,


carleyjmartin:
Hi Francis, I had a few errors using the code you supplied. I assume that the line `Coords` is supposed to read `'coords': Coords.GEOGRAPHIC` and I've removed the `cmap:viridis` line as this was throwing an error, so the code used to test was:
```python
import pydarn
import pydarnio

from datetime import datetime
from pydarn import Coords

sps_fitacf_file = '/Users/carley/Documents/data/20200101.0001.00.inv.fitacf'

sps_fitacf_reader = pydarnio.SDarnRead(sps_fitacf_file)
sps_fitacf_data = sps_fitacf_reader.read_fitacf()

fig = plt.figure(figsize=(12,12))
kwargs1 = {
'boundary': True,
'zmax': 200,
'zmin': -200,
'groundscatter': False,
'scan_index':22,
'colorbar_label':'Power [dB]',
'line_color':'red',
'radar_label': True,
'parameter': 'p_l',
'coords': Coords.GEOGRAPHIC
}

ax, beam_corners_lats, beam_corners_lons, scan, grndsct = \
    pydarn.Fan.plot_fan(sps_fitacf_data, colorbar=True, **kwargs1)
    
plt.show()
```

I think the kwargs probably need tweaking as it's plotting p_l not velocity, but this is the result:
Screen Shot 2022-07-04 at 9 32 25 AM
Does this look right?

There are some old hardware files on this branch too instead of the link to the hardware repo, I had to replace them to get this to work, so they will need to be removed so we don't merge any hdw files into develop.

The y0inx line was originally used as the inner few gates end up at zero/infinity when using the `'range_estimation': RangeEstimation.GSMR` option and then cause a bit of havoc later down the line when plotting, so they were removed here.
When testing with this option the code does break as there are these non-finite values (error below), so I think we might need a different work around than removing the line. If the aim is to just return the values for all gates/beams then this small bit of code to return the values could be it's own method then it doesn't have to affect the plotting functions or it could be an option to choose when calling the Coords function? If you need any help with that or more explanation let me know! :) 

Is this only required for the Geographic coordinates, as the y0inx line is also in the AACGM function too?

```
  File "/Users/carley/Documents/testing_code/test_music.py", line 30, in <module>
    pydarn.Fan.plot_fan(sps_fitacf_data, colorbar=True, **kwargs1)
  File "/Users/carley/Documents/testing_code/venv-music1/lib/python3.9/site-packages/pydarn/plotting/fan.py", line 320, in plot_fan
    ax.pcolormesh(thetas, rs,
  File "/Users/carley/Documents/testing_code/venv-music1/lib/python3.9/site-packages/matplotlib/__init__.py", line 1412, in inner
    return func(ax, *map(sanitize_sequence, args), **kwargs)
  File "/Users/carley/Documents/testing_code/venv-music1/lib/python3.9/site-packages/matplotlib/axes/_axes.py", line 6058, in pcolormesh
    X, Y, C, shading = self._pcolorargs('pcolormesh', *args,
  File "/Users/carley/Documents/testing_code/venv-music1/lib/python3.9/site-packages/matplotlib/axes/_axes.py", line 5533, in _pcolorargs
    raise ValueError(
ValueError: x and y arguments to pcolormesh cannot have non-finite values or be of type numpy.ma.core.MaskedArray with masked values

```

francistee26:
Yeah, sorry about that.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

Issue #268, opened 2022-07-07
BUG: Issue plotting Infinity for GSMR Range Estimation - Help needed for further dev in MUSIC code


carleyjmartin:
# BUG 

For further development, it is requested that we remove the line that removes infinite values from 
`coordinates.py` 
```python
y0inx = np.min(np.where(np.isfinite(beam_corners_lats[:,0]))[0])
```
As these values cause issues with plotting further down the line. 

We need a solution where this line is removed/amended and the size of the `return beam_corners_lats, beam_corners_lons` is the size of the fan, but does not affect the plotting later on. Any ideas welcome!

@w2naf @francistee26 please add anything that you think would be helpful to know!

## Priority

- [ ] Minor - typo, incorrect naming, fails on specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation 
- [X] Urgent - failure in usage, large misrepresentation

Urgent only as it is required for further development, and is a failure in usage if removed. 

## Information
**python version**:
**OS**:
**matplotlib version**:

## Example of the bug 

Error output when plotting a fan plot with `'range_estimation': RangeEstimation.GSMR` with the line in the code removed:
```
Traceback (most recent call last):
  File "/Users/carley/Documents/testing_code/test_music.py", line 30, in <module>
    pydarn.Fan.plot_fan(sps_fitacf_data, colorbar=True, **kwargs1)
  File "/Users/carley/Documents/testing_code/venv-music1/lib/python3.9/site-packages/pydarn/plotting/fan.py", line 320, in plot_fan
    ax.pcolormesh(thetas, rs,
  File "/Users/carley/Documents/testing_code/venv-music1/lib/python3.9/site-packages/matplotlib/__init__.py", line 1412, in inner
    return func(ax, *map(sanitize_sequence, args), **kwargs)
  File "/Users/carley/Documents/testing_code/venv-music1/lib/python3.9/site-packages/matplotlib/axes/_axes.py", line 6058, in pcolormesh
    X, Y, C, shading = self._pcolorargs('pcolormesh', *args,
  File "/Users/carley/Documents/testing_code/venv-music1/lib/python3.9/site-packages/matplotlib/axes/_axes.py", line 5533, in _pcolorargs
    raise ValueError(
ValueError: x and y arguments to pcolormesh cannot have non-finite values or be of type numpy.ma.core.MaskedArray with masked values
```
### Attempts

No attempts made.

## Potential Solution(s)

[...]

PR #270, opened 2022-07-08
HOT fix for both music code and pydarn when needing either full_array or limited array of lats and lons

francistee26:
# Scope 

HOT fix for both music code and pydarn when needing either full_array or limited array of lats and lons in geo_coordinates function.

## Approval

**Number of approvals:** 3

## Test
```
import pydarn
import pydarnio

from datetime import datetime
from pydarn import Coords

bks_fitacf_file = '/home/fran/code/SuperdarnW3usr/ForGitRepo/2010/fitacf/bks/20101119.1401.00.bks.fitacf'

bks_fitacf_reader = pydarnio.SDarnRead(bks_fitacf_file)
bks_fitacf_data = bks_fitacf_reader.read_fitacf()

fig = plt.figure(figsize=(12,12))
kwargs1 = {
'boundary': True,
'zmax': 200,
'zmin': -200,
'groundscatter': False,
'scan_index':22,
'colorbar_label':'Velocity m/s',
'line_color':'blue',
'radar_label': True,
'parameter': 'v',
'full_array':False
}

ax, beam_corners_lats, beam_corners_lons, scan, grndsct = \
    pydarn.Fan.plot_fan(bks_fitacf_data, colorbar=True, **kwargs1) 
```

### DIFF
--- a/pydarn/plotting/fan.py
+++ b/pydarn/plotting/fan.py
@@ -74,7 +74,7 @@ class Fan():
                  colorbar_label: str = '', title: bool = True,
                  boundary: bool = True, projs: Projs = Projs.POLAR,
                  coords: Coords = Coords.AACGM_MLT,
-                 channel: int = 'all', **kwargs):
+                 channel: int = 'all',full_array:bool = False, **kwargs):
         """
         Plots a radar's Field Of View (FOV) fan plot for the given data and
         scan number
@@ -138,6 +138,9 @@ class Fan():
                 integer indicating which channel to plot or 'all' to
                 plot all channels
                 Default: 'all'
+            full_array: bool 
+                boolen indicating to return the full array of both lons and lats coordinates
+                Default: false
             kwargs: key = value
                 Additional keyword arguments to be used in projection plotting
                 and plot_fov for possible keywords, see: projections.axis_polar
@@ -235,7 +238,7 @@ class Fan():
         beam_corners_lats, beam_corners_lons =\
                 coords(stid=dmap_data[0]['stid'],
                        rsep=rsep, frang=frang,
-                       gates=ranges, date=date,
+                       gates=ranges, date=date,full_array=full_array,
                        **kwargs)
 
         fan_shape = beam_corners_lons.shape
--- a/pydarn/utils/coordinates.py
+++ b/pydarn/utils/coordinates.py
@@ -29,7 +29,7 @@ from pydarn import (geocentric_coordinates, SuperDARNRadars, RangeEstimation,
 
 
 def geo_coordinates(stid: int, beams: int = None,
-                    gates: tuple = None, **kwargs):
+                    gates: tuple = None, full_array: bool = False,**kwargs):
     """
     geographic_coordinates calculates the geographic coordinate for a given
     set of gates and beams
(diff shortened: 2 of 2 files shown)


carleyjmartin (commented):
Hi Francis, 

I'm assuming again that you want this to be working when you include `'range_estimation':pydarn.RangeEstimation.GSMR` as this is the only time that the beam_corner_lats and beam_corner_lons is actually modified to be smaller. 

As it stands, the test code provided works as it isn't using `full_array:True` or the GSMR range estimation, but adding them in and testing for `full_array:True`. I also just realized that the if statement is only present in the geo_coordinates method, so to test we need to add `'coords':pydarn.Coords.GEOGRAPHIC, 'projs':pydarn.Projs.GEO` as well. When including this I get the same error as before as we're trying to plot with infinities, this break comes before the return so we do not return the values of the beam corners arrays.
Do you need this code to also be included in the AACGM function too to work for both coordinate systems, or it it something solely used in geographic coordinates?

I think that to get this version working you will need to call geo_coordinates directly, not through a plotting function to get it to return the size array you want. I don't know what you need to do after you have the array, but if it's plotting a fan we might need to think about my next idea below.

I was thinking that maybe changing the infinity values to a NaN instead might solve the plotting issues and then it will always return the correct size and will be dealt with easier in the plotting. I can attempt a test of doing it this way (and the above way) in the next week or so to see if it's an easier solution/if it works. 

The suggestions in the review are for PEP8 styling and some additional info on the key word so users don't try to plot with it, but as it stands I'm not getting the values for the beams as the code is breaking before the return. 

The code I used is as follows:
```python
mport pydarn
import pydarnio

from datetime import datetime
from pydarn import Coords, RangeEstimation

sps_fitacf_file = '/Users/carley/Documents/data/20200101.0001.00.inv.fitacf'

sps_fitacf_reader = pydarnio.SDarnRead(sps_fitacf_file)
sps_fitacf_data = sps_fitacf_reader.read_fitacf()

fig = plt.figure(figsize=(12,12))
kwargs1 = {
'boundary': True,
'zmax': 200,
'zmin': -200,
'groundscatter': False,
'scan_index':22,
'colorbar_label':'Velocity m/s',
'line_color':'blue',
'radar_label': True,
'parameter': 'v',
'full_array':True,
'range_estimation':pydarn.RangeEstimation.GSMR,
'coords':pydarn.Coords.GEOGRAPHIC,
'projs':pydarn.Projs.GEO
}

ax, beam_corners_lats, beam_corners_lons, scan, grndsct = \
    pydarn.Fan.plot_fan(sps_fitacf_data, colorbar=True, **kwargs1) 

print(beam_corners_lats.size)
```

carleyjmartin:
So I've had a quick look at what to do with this. We can't keep the NaN's in the array and pass it to the fan plotting method because pcolormesh does not allow anything but actual finite values in the X or Y inputs so that's not an option. If you are wanting to plot these NaN values I don't think there is a solution to do that as a fan plot at least, other than redo the GSMR equations to not remove the first several range gates (which I'm assuming is also something that is not possible)?

Let me know what you actually need as the input for the music algorithm, because you can't really do anything with NaN values for a lat-lon position anyway, I *think* what you might want is the size of the original array? It might be better to return the original size i.e. 16 x 75 (beams x range gates)? If you want the actual NaN values and not plot anything I think the best option would be to make your own separate method that returns what you want as the geo_coordinates method is wrapped up in the Coords enum method (bottom of coordinates.py).

Again, the beam_corners_lats and beam_corners_lons arrays are only reduced in size when using the ground scatter mapped range estimation, so for the testing code provided the code will not break as there are no NaNs in the arrays to remove (i.e. beam_corners_lats is the same size as beam_corners_lats[y0inx:] for slant range). 

So sorry, essentially both my ideas will not work with the current code.
I did make an issue last week to ask the community if they had any ideas to help fix this #268 but I don't think it's had any traction yet.

Issue #276, opened 2022-08-16
BUG: Plotting convection maps too slow


carleyjmartin:
# BUG 

I noticed whilst trying to mass produce convection plots that they take almost 20 seconds to make one plot and this isn't okay.
After doing some timings, I realised it was because we are iterating over each STID for FOV which is taking 15 seconds (including cartopy call, without cartopy it is 13 seconds). Unless the user wants to plot the FOV outlines, I think we can get away with just calling this once with an option to iterate? I'm going to give this a try anyway for personal use and if it works with all options I'll put it in a PR, but let me know if you have any other ides to save time! The main other time drain is calculating the contours, and I don't think we can reduce that so much.

## Priority

- [x] Minor - typo, incorrect naming, fails on specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

## Information
**python version**: 3.9
**OS**: Mac Monterey 12.4
**matplotlib version**: 3.5.3

## Example of the bug 
I stuck time.time() around all the sections:

Finding Record: 1.3828277587890625e-05
**FOV calls: 15.00069785118103**
Shifting MLT/AACGMv2: 0.0002579689025878906
Fitted velocities: 0.005090951919555664
Rotate vectors: 0.0002613067626953125
Plotting Vectors: 0.03212380409240723
Contours: 5.892899036407471
HMB: 0.0008921623229980469
Title: 0.00016617774963378906
Info: 0.00011301040649414062
IMF: 0.011370182037353516
Total: 20.944074153900146

## Potential Solution(s)

Use boundary True False to iterate or not to show FOV, then maybe only call once if false?

PR #277, opened 2022-08-16
EHN: Speed up common use of convection maps

carleyjmartin:
# Scope 

This PR speeds up the plotting of default convection maps when the user doesn't want FOV or radar locations (which is default).
If boundary (FOV) = False and radar_location=False then the code will now skip calling the plot_FOV function and call a basic projection to get the axes. If either are True, then it will call each FOV to plot individually as before. 

This removes ~13 seconds from the plotting time of one convection map, reducing the total time taken by 2/3, yay!
Using the coastline=True option will increase time taken, as its calling some Cartopy functions and then is converted to MLT, by 2 seconds. If coastlines=False, then the whole axis plotting is now only 0.1 seconds, making a vast improvement on speed. 

The remaining time taken (~6 seconds) is plotting the contours, specifically calling the special.lpmn function ~7000 times to calculate the Legendre functions - this function can only take scalar values so cannot be vectorized to speed it up unfortunately so I don't see an easy way to save time here.

Also added a warning as we are not actively developing maps with other projections.

**issue:** #276 

## Approval

**Number of approvals:** It's a simple if loop so probably 1 good test and review is okay. 

## Test

**matplotlib version**: 3.5.3
```python
import pydarn
import datetime as dt

# Requires Cartopy installation for coastlines - just remove coastlines for non-cartopy
            lowlat=50, color_vectors=False,
            contour_fill=True, coastline=True)
plt.show()
```
Produces:
Screen Shot 2022-08-16 at 1 03 07 PM

**Note: double check that hemisphere is being read into projs for southern data to work too**

**Note: check grid plots for same issue**

### DIFF
--- a/pydarn/plotting/maps.py
+++ b/pydarn/plotting/maps.py
@@ -9,6 +9,7 @@
 # 2022-04-01: CJM - Bug fix for lon shifting to MLT
 # 2022-04-28: CJM - Added option to have single color vectors with reference
 #                   vector
+# 2022-08-15: CJM - Removed plot_FOV call for default uses
 #
 # Disclaimer:
 # pyDARN is under the LGPL v3 license found in the root directory LICENSE.md
--- a/pydarn/plotting/maps.py
+++ b/pydarn/plotting/maps.py
@@ -40,7 +40,7 @@ import aacgmv2
 
 from pydarn import (PyDARNColormaps, plot_exceptions,
                     standard_warning_format, Re, Hemisphere,
-                    time2datetime, find_record, Fan, MapParams)
+                    time2datetime, find_record, Fan, Projs, MapParams)
 
 warnings.formatwarning = standard_warning_format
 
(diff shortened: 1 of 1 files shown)


SEC-VT:
I found a way to vectorize the Legendre polynomial function from scipy:
```python
def vec_lpmn(m, n, x):
    return special.lpmn(m, n, x)[0]
vec_lpmn = np.vectorize(vec_lpmn, signature='(),(),()->(k,k)')
```
When creating numpy ufunc that return nonscalar outputs, you have to modify the [ufunc signature](https://numpy.org/doc/stable/reference/c-api/generalized-ufuncs.html#details-of-signature).

In my testing this doesn't speed up single plot generation (since instantiating a matplotlib axes itself takes about 4s), but there may be gains when updating the data on subsequent plots.

Elsewise, this PR looks good to me.

carleyjmartin:
Thanks @SEC-VT !
If you think the PR is good, I think it only really needs 1 approval so feel free to merge. 
(We can stick the vectorizing in another PR when it's been put in later, smaller PRs are far easier to get merged, it might be good to make an issue on this actually)

Issue #284, opened 2022-09-28
BUG: Matplotlib depreciation of mpl_toolkits.axes_grid


carleyjmartin:
# BUG 

The new release of matplotlib 3.6.0 does not allow for us to use mpl_toolkits.axes_grid in the maps.py module any more. 

```
Traceback (most recent call last):
  File "test.py", line 1, in <module>
    import pydarn
  File "pydarn/__init__.py", line 68, in <module>
    from .plotting.maps import Maps
  File "plotting/maps.py", line 32, in <module>
    from mpl_toolkits.axes_grid.inset_locator import InsetPosition
ModuleNotFoundError: No module named 'mpl_toolkits.axes_grid'
```

## Priority

- [ ] Minor - typo, incorrect naming, fails on specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation 
- [x] Urgent - failure in usage, large misrepresentation

## Information
**python version**: 3.9
**OS**: Mac Monterey 12.6
**matplotlib version**: **3.6.0**

## Example of the bug 

This happens on import so literally any piece of code, even not including maps will break.

## Potential Bug Location 

This is used in maps.py to position the IMF clock. 

## Potential Solution(s)

Not a clue, I remember that positioning the clock and keeping all its parts relative to each other was not possible by other methods so this might be difficult.

PR #285, opened 2022-09-29
FIX: mpl axes_grid depreciation fix

carleyjmartin:
# Scope 

This fixes the depreciation of the axes_grid module of matplotlib -appears to have just changed to axes_grid1.

**issue:** #284 

## Approval

**Number of approvals:** 1 (1 liner)

## Test

**matplotlib version**: tested on 3.6.0 (new version that won't work before), 3.5.3 and 3.3.4 (lowest compatible version with pyDARN)

imports correctly:
`import pydarn`

You can check that the map plots are still working with:
```python
import pydarn
map_file = "/Users/carley/Documents/data/maps/20170113.s.map"
            lowlat=30, color_vectors=False,
            contour_fill=True, coastline=True)
plt.show()
```

### DIFF
--- a/pydarn/plotting/maps.py
+++ b/pydarn/plotting/maps.py
@@ -32,7 +32,7 @@ import warnings
 
 from enum import Enum
 from matplotlib import ticker, cm, colors
-from mpl_toolkits.axes_grid.inset_locator import InsetPosition
+from mpl_toolkits.axes_grid1.inset_locator import InsetPosition
 from scipy import special
 from typing import List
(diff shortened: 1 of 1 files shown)


KhanalKrishna:
Matplotlib version: 3.6.1

**Test script:**
import pydarn
map_file = "20110128.north.map2"
            lowlat=30, color_vectors=False,
            contour_fill=True, coastline=True)
plt.show()

image

This looks good to me.

PR #297, opened 2023-02-03
FIX: hdw repo not installing

### LINKED ISSUE #292: BUG: hdw repo not being cloned on install
# BUG 

I've had a few reports of the hdw repository not being downloaded on install. 
I can recreate this error, but only when trying to install system wide.

Until a solution is made I recommend that if you find this error you can try:
1. Installing and running pyDARN programs in a virtual environment https://docs.python.org/3/library/venv.html
2. Cloning the hdw repository in the `utils` directory of wherever pyDARN has installed https://github.com/SuperDARN/hdw

If you do find this error, please comment with the traceback, and so that I know if it's widespread or os dependent. 

## Priority

- [ ] Minor - typo, incorrect naming, fails on specific uncommon situation

## Information
**python version**: 3.10
**OS**: MacOS Monterey
**matplotlib version**: NA

## Example of the bug 

Occurs on `import pydarn` on new install. 

### Attempts

*What have you tried already to fix it?*
NA

*What have you tried to get around it?*
See above for solutions in the meantime. 

## Potential Bug Location 

On install during hdw repo cloning. 

## Potential Solution(s)

Any ideas welcome.

carleyjmartin:
# Scope 

This PR is an attempt to fix the hdw repo not installing on new pip installations. I *think* this will fix it, but I will need to try it out on testpypi first I think. In fixing this, another depreciation warning for pip is showing saying that setup.py install is going to be depreciated *this may or may not be an issue for us, it's very unclear what is happening with that*.

**issue:** to close #292 

## Approval

**Number of approvals:** 2

## Test

**matplotlib version**: NA

I will update here when I try to put it on testpypi, in the meantime you can try doing 
`pip3 install git+https://github.com/superdarn/pydarn@fix/hdw-repo`
Installing this way didn't have a issue with the hdw repo before so it should just work anyway.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,7 +22,7 @@ import sys
 import warnings
 
 from os import path
-from setuptools import setup, find_packages
+from setuptools import setup, find_namespace_packages
 from setuptools.command.install import install, orig
 from setuptools.command.develop import develop
 from glob import glob
@@ -83,7 +83,7 @@ setup(
         'Programming Language :: Python :: 3.6',
         'Programming Language :: Python :: 3.7'],
     python_requires='>=3.6',
-    packages=find_packages(['pydarn/utils/hdw'],exclude=['docs', 'test']),
+    packages=find_namespace_packages(exclude=['docs', 'test']),
     author="SuperDARN Data Visualization Working Group",
     # used to import the logging config file into pydarn.
     include_package_data=True,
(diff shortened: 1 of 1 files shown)


carleyjmartin:
Fix coming that actually works!

PR #305, opened 2023-03-07
Return map potential data from Map Plots

shibaji7:
# Map potential data

Return fitted potential / velocity / e-field datasets from Map potential data plots. This enhancement will help to share the fitted datasets for event analysis in various conjunction studies using SuperDARN. @bharatreddy @Shirling-VT @carleyjmartin 

## Approval

**Number of approvals:** 2 (testing and code review where possible)

### DIFF
--- a/pydarn/plotting/maps.py
+++ b/pydarn/plotting/maps.py
@@ -158,6 +158,19 @@ class Maps():
                 uses the parameters for plot_fov and projections.axis
 
         """
+        #####
+        # Adding return statements for different level processed data: 
+        #              1. potential, 2. velocity (model/vectored), 3. e-field
+        #####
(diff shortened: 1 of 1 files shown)


bharatreddy:
Why change the "plot_mapdata" function (https://github.com/SuperDARN/pydarn/blob/0e459a3808d60e8e810d72526bd70d6f62c94d4d/pydarn/plotting/maps.py#L67)? I personally feel it might be better to change the "calculated_fitted_velocities" function (https://github.com/SuperDARN/pydarn/blob/aabbb70b5e71becfdaddbc63db142f42d729eb16/pydarn/plotting/maps.py#L455) instead. Any advantages?

PR #335, opened 2023-06-15
expose potential vmin vmax

w2naf:
# Scope 

**issue:** This pull request lets the user manually specify the vmin and vmax values for the map potential contour colorbars.

## Approval

**Number of approvals:** 2

## Test

**matplotlib version**: 3.7.1

*provide a list or code to test*

```
import bz2
import pydarn
import pydarnio

map_file = "20220202.north.map2.bz2"
with bz2.open(map_file) as fp:
    dmap_stream = fp.read()

reader    = pydarnio.SDarnRead(dmap_stream, True)
map_data  = reader.read_map()

pydarn.Maps.plot_mapdata(map_data, record=150, 
                         lowlat=40, colorbar_label='Velocity m/s',
                         contour_fill = True,
                         contour_fill_cmap= 'RdBu',
                         contour_colorbar = True,
                         contour_colorbar_label='Potential (kV)',
                         pot_minmax_color = 'r',
                         pot_vmin = -30,
                         pot_vmax = 30,
                         map_info=True, imf_dial=True, hmb=True)
                         
plt.show()
```

  - copyrighted (if applicable)

### DIFF
--- a/pydarn/plotting/maps.py
+++ b/pydarn/plotting/maps.py
@@ -75,6 +75,8 @@ class Maps():
                      hmb: bool = True, boundary: bool = False,
                      radar_location: bool = False, map_info: bool = True,
                      imf_dial: bool = True, reference_vector: int = 500,
+                     pot_vmin: float = None,
+                     pot_vmax: float = None,
                      projs: Projs = Projs.POLAR, **kwargs):
         """
         Plots convection maps data points and vectors
@@ -150,6 +152,12 @@ class Maps():
                 Vector can be turned off by setting this value to False or 0.
                 Will not plot for power or spectral width options.
                 Default: 500 (vector plotted)
+            pot_vmin: float
+                Minumum value of color map used for potential contours.
+                If None, defaults to -abs(pot_arr).max()
+            pot_vmax: float
+                Maximum value of color map used for potential contours.
+                If None, defaults to abs(pot_arr).max()
             projs: Enum
                 choice of projection for plot
                 default: Projs.POLAR (polar projection)
@@ -396,6 +404,7 @@ class Maps():
         cls.plot_potential_contours(fit_coefficient, lat_min, date, ax,
                                     lat_shift=lat_shift, lon_shift=lon_shift,
                                     fit_order=fit_order, hemisphere=hemisphere,
+                                    pot_vmin=pot_vmin,pot_vmax=pot_vmax,
                                     **kwargs)
 
         if hmb is True:
@@ -896,7 +905,10 @@ class Maps():
                                 contour_colorbar: bool = True,
                                 contour_fill_cmap: str = 'RdBu',
                                 contour_colorbar_label: str = 'Potential (kV)',
-                                pot_minmax_color: str = 'k', **kwargs):
+                                pot_minmax_color: str = 'k', 
+                                pot_vmin: float = None,
+                                pot_vmax: float = None,
+                                **kwargs):
         # TODO: No evaluation of coordinate system made! May need if in
         # plotting to plot in radians/geo ect.
         '''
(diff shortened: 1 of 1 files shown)


carleyjmartin:
Hey @w2naf thanks for checking this out. I think the option to set the contour colorbar min/max might be a bit more complicated than just setting the min max values as it is right now. We set the contour level positions as default from -100 to 100 (line 1003), or the user can set their own using the keyword `contour_levels`. This is still fixed regardless of the colorbar pot_zmin pot_zmax set at the moment, in the example below you can see that the colours in the bar don't really correspond to the color in the contours on the plot as the contour plot is still taking -100/100 as the min max values. Left is pot_zmin/max=+-10, right is +-100, but the contours are always +-100 regardless.

Screenshot 2023-06-16 at 8 57 40 AM

I think the solution here might be to amend the contour_levels if a pot_zmin/pot_zmax is set by the user in/around lines 1002-1006. 
...
*after thinking about it a bit more* 
....
Now I think about it more, setting the `contour_levels` keyword itself will set the min max of the colorbar automatically for you. E.G. if you set `contour_levels=[-30,-20,-10,-1,1,10,20,30]` you will get the same result as the example in your pr comment. I think we went with contour level earlier on as you can set the minmax of the contours, but if you don't fix around zero you end up with a weird half colored in map that has a contour at 0.

Let me know what you think, I'm pretty sure we can figure something out that's better than how the contours are done right now.

w2naf:
Hi, @carleyjmartin. Thanks for your response!

Perhaps we can make the behavior like this:
- contour_levels automatically adjusts for pot_zmin/pot_zmax if pot_zmin/pot_zmax is specified and contour_levels=None
- pot_zmin/pot_zmax automatically conforms to contour_levels if contour_levels is specified but pot_zmin/pot_zmax are set to None.
- If all parameters are specified, follow the user inputs exactly.
- If no parameters are specified, have the system autoscale as it does now.

I'm curious, why do you use zmin/zmax instead of vmin/vmax? Matplotlib uses vmin/vmax for all of its cmap calls.

Thank you again!

carleyjmartin:
Sounds good to me!

Honestly, I was probably following the convention set with fan plots which used zmin/zmax first. Which I think may have been switched so users don't automatically associated this function with 'v' for velocity

w2naf:
Thanks, @carleyjmartin! That all sounds good. I can work on coding this up, but it may take some time because of other higher priorities right now. Let me know if you might have a chance of coding it up instead.

carleyjmartin:
Okay, I think I got something working. I've added a keyword `contour_spacing` too to give the user a lot more control over where they want to see the levels. By default `pot_zmin` and `pot_zmax` are -50 and 50, and the default spacing turns out to be 5 with these defaults. The default spacing is worked out between the zmin and zmax to have 20 levels on the plot.  I've kept the 'z' instead of 'v' too to avoid any issue with user thinking they're setting the velocity color bar min and max. I've removed the `pot_vmin` and `pot_vmax` from the `plot_mapdata` method as the keywords will go through to the contour plotting method using `**kwargs`.

This also fixes a bug where the top/bottom contours always matched the top/bottom colours of the color bar regardless of the actual value. So you can still set your own levels in a list, if not you can set a min and max and the code will work out the levels for you with/without a given spacing. Contour positioning by any method works for filled or unfilled contours.

Examples:
Using only default values:
```python
import pydarn

map_file = "/Users/carley/Documents/data/maps/20190421.s.map2"
pydarn.Maps.plot_mapdata(map_data, record=150, 
                         lowlat=60,
                         color_vectors=False, contour_fill=True)
plt.show()
```
Screenshot 2023-06-30 at 1 28 32 PM


Setting your own min and max and spacing:
```python
pydarn.Maps.plot_mapdata(map_data, record=150, 
                         lowlat=60,
                         color_vectors=False, contour_fill=True,
                         pot_zmin = -10, pot_zmax = 10, contour_spacing=1)
plt.show()
```
Screenshot 2023-06-30 at 1 29 27 PM



Complete side note, but I used the GitHub CLI today for the first time to develop on this branch as it's a fork and would absolutely recommend for anyone else, only thing is it pushed to a new pydarn branch instead of your fork, will go make a PR and link it here!

Issue #334, opened 2023-06-14
New Feature: Time Series Plots for Scalar Map Data


carleyjmartin:
# New Feature

**Name**: plot_time_series()

**module**:  *map.py* 

**package**: plotting

Plot a time series of data from map files (opposed to the existing fitacf version)

## Description

This feature will be able to plot data over time for things made in map files (hmb, cpcp +)

### pyDARN Checklist 

-  [x] Does this fit within pyDARN's scope? 
-  [ ] Is this a minor change i.e., new small utility function?
-  [x] Is this a major change i.e., new plotting function?
-  [ ] Do you need help developing it?

PR #336, opened 2023-06-19
Feature: Time Series Plots from Map files

PrestonXPitzer:
# Scope 

Adds a method that graphs some map file parameters in time. Currently added are the number of vectors, IMF By/Bz magnitudes, KP index, and maximum HMB latitude. 

I'm having issues with getting the correct record index from an input datetime object, so I left it unimplemented for now. 

Implements issue #334 

## Approval

0 approvals, needs 2

## Test

**matplotlib version**:  3.7.1

```python

import bz2 
import pydarn
import datetime as dt

file_path = "example_data/20150310.north.map2.bz2"
data = pydarn.SuperDARNRead().read_dmap(file_path)
pydarn.Maps.plot_time_series(data, parameter=pydarn.Maps.TimeSeriesParams.NUM_VECTORS,
                             start_record=0, end_record=100)
plt.show()

```
example_forpr

### DIFF
--- a/pydarn/plotting/maps.py
+++ b/pydarn/plotting/maps.py
@@ -37,13 +37,14 @@ from mpl_toolkits.axes_grid1.inset_locator import InsetPosition
 from scipy import special
 from typing import List
 
+
 # Third party libraries
 import aacgmv2
 
 from pydarn import (PyDARNColormaps, plot_exceptions,
                     standard_warning_format, Re, Hemisphere,
-                    time2datetime, find_record, Fan, Projs, MapParams)
-
+                    time2datetime, find_record, Fan, Projs, 
+                    MapParams, TimeSeriesParams)
 warnings.formatwarning = standard_warning_format
 
 
@@ -1030,39 +1031,57 @@ class Maps():
                     color=pot_minmax_color, zorder=5.0)
         plt.scatter(np.radians(min_mlon), min_mlat, marker='_', s=70,
                     color=pot_minmax_color, zorder=5.0)
+    @classmethod
+    def find_map_record(cls,dmap_data: List[dict], start_time: dt.datetime):
+        """
+        looks through the data from a given map file and returns the record number nearest the 
+        passed datetime object
         
+        Parameters
+        -----------
+        dmap_data: List[dict]
+            the data to look through
+        start_time: datetime
+            the time to find the nearest record number to
+
+        Returns
+        -------
+        Returns the closet record to the passed time
+        """
+        #recursively identify the index of the list where the time is closest to the passed time
+        def find_nearest_time(dmap_data: List[dict], start_time: dt.datetime, start_index: int, end_index: int):
+            #base case
+            if start_index == end_index:
+                return start_index
+            #recursive case
+            else:
+                #find the middle index
+                mid_index = int((start_index + end_index)/2)
+                #if the time at the middle index is greater than the passed time
+                if time2datetime(dmap_data[mid_index]) > start_time:
+                    #search the lower half of the list
+                    return find_nearest_time(dmap_data, start_time, start_index, mid_index)
+                #if the time at the middle index is less than the passed time
+                elif time2datetime(dmap_data[mid_index]) < start_time:
+                    #search the upper half of the list
+                    return find_nearest_time(dmap_data, start_time, mid_index + 1, end_index)
+                #if the time at the middle index is equal to the passed time
+                else:
+                    #return the middle index
+                    return mid_index
+        
+        return find_nearest_time(dmap_data, start_time, 0, len(dmap_data) - 1)
     
-
-    class TimeSeriesParams(Enum):
-        """
-        Enum class to hold the parameters that can be plotted
-        Members
-        ------------
-        NUM_VECTORS:
-        the number of aviliable vectors in the map
-        IMF_BY:
-        The magnitude of the IMF By component in nT
-        IMF_BZ:
-        The magnitude of the IMF Bz component in nT
-        KP:
-        The Kp index
-        HMB_MAX:
-        The farthest equatorward extent of the HMB
-
-        """
-        NUM_VECTORS = 'num_vectors'
-        IMF_BY = 'imf_by'
-        IMF_BZ = 'imf_bz'
-        KP = 'kp'
-        HMB_MAX = 'hmb_max'     
         
+
+
     @classmethod
     def plot_time_series(cls, dmap_data: List[dict],
                      parameter: Enum = TimeSeriesParams.NUM_VECTORS,
                      start_record: int = 0,end_record: int = 1, start_time: dt.datetime = None,
-                     end_time: dt.datetime = None):
+                     end_time: dt.datetime = None, ax=None,**kwargs):
         '''
-        Plot time series of various map parameters
+        Plot time series of various map parameters  
         
         Params
         ----------
@@ -1076,22 +1095,28 @@ class Maps():
             Start time of the data
         end_time: dt.datetime
             End time of the data
+        ax:
+            matplotlib axis
+        kwargs:
+            keyword arguments to be passed to the plotting function
+        
             '''
-        #create a new matplotlib figure
-        fig = plt.figure()
-        #if end_record is -1, set it to the last record
-        if end_record == -1:
+        
+        #if no time objects are passed & no start/end record is passed
+        #then plot all availaible data
+        if start_record == 0 and end_record == 1 and start_time is None and end_time is None:
+            start_record = 0
             end_record = len(dmap_data)-1
+        
         #determine the start and end record
         if start_time is not None and end_time is not None:
-            raise NotImplementedError('use of datetime objects is not yet implemeneted')
-            #start_record = find_record(dmap_data, start_time) #TODO: Fix the error that gets thrown when using this
-            #end_record = find_record(dmap_data, end_time)
+            start_record = cls.find_map_record(dmap_data, start_time)
+            end_record = cls.find_map_record(dmap_data, end_time)
         else:
             start_time = time2datetime(dmap_data[start_record])
             end_time = time2datetime(dmap_data[end_record])
         #based on the parameter, plot the data
-        if parameter == cls.TimeSeriesParams.NUM_VECTORS:
+        if parameter == TimeSeriesParams.NUM_VECTORS:
             datalist = []
             timelist = []
             for records in range(start_record, end_record):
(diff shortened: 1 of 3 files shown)


pydarn/plotting/maps.py
```diff
@@ -1030,3 +1030,117 @@ def plot_potential_contours(cls, fit_coefficient: list, lat_min: list,
...
+            End time of the data
+            '''
+        #create a new matplotlib figure
+        fig = plt.figure()
+        #if end_record is -1, set it to the last record
+        if end_record == -1:
+            end_record = len(dmap_data)-1
+        #determine the start and end record
+        if start_time is not None and end_time is not None:
+            raise NotImplementedError('use of datetime objects is not yet implemeneted')
```
carleyjmartin, line 1087: So I removed this and uncommented the two lines below and I can get find_record to work if I use:

```python
file_path = "/Users/carley/Documents/data/maps/20190421.s.map2"
data = pydarn.SuperDARNRead().read_dmap(file_path)
start_time = dt.datetime(2019,4,21,11,0)
end_time = dt.datetime(2019,4,21,13,0)
pydarn.Maps.plot_time_series(data, parameter=pydarn.Maps.TimeSeriesParams.HMB_MAX,
                             start_time=start_time, end_time=end_time)
plt.show()
```

I think there was an old issue that didn't allow you to put seconds in there, because it would break? But either way it works for me using these date times. What is the error you got, we can maybe do a find_map_record or something? (or add an option in find_records to find nearest second/minute/2minutes then it's useful for all data types, this is probably a different issue to fix+PR though? Can talk about it friday at the meeting)

As map files have data every two mins, if I use any other number for seconds than 0 I get:
```
  File "/Users/carley/Documents/testing_code/venv-timemap/pydarn/../../test_maptimeseries.py", line 9, in <module>
    pydarn.Maps.plot_time_series(data, parameter=pydarn.Maps.TimeSeriesParams.HMB_MAX,
  File "/Users/carley/Documents/testing_code/venv-timemap/lib/python3.10/site-packages/pydarn/plotting/maps.py", line 1088, in plot_time_series
    start_record = find_record(dmap_data, start_time) #TODO: Fix the error that gets thrown when using this
  File "/Users/carley/Documents/testing_code/venv-timemap/lib/python3.10/site-packages/pydarn/utils/plotting.py", line 67, in find_record
    raise plot_exceptions.NoDataFoundError(parameter, start_time=start_time)
NameError: name 'parameter' is not defined
```
Which I guess is expected because we don't read in any parameter to the function.... so this an error in the exception it raises, and a different problem to solve. but essentially, it's looking for data at a time where data doesn't exist. I think the solution here is to make a new find_map_record that checks nearest minute, rather than nearest second?

pydarn/plotting/maps.py
```diff
@@ -1030,3 +1030,117 @@ def plot_potential_contours(cls, fit_coefficient: list, lat_min: list,
...
+            Start time of the data
+        end_time: dt.datetime
+            End time of the data
+            '''
+        #create a new matplotlib figure
+        fig = plt.figure()
+        #if end_record is -1, set it to the last record
+        if end_record == -1:
+            end_record = len(dmap_data)-1
+        #determine the start and end record
```
carleyjmartin, line 1085: I think maybe for clarity, we should do the same as rtp.py time series plotting does where if the user doesn't put in a start and end time, then we do the whole time frame of data read in (in __determine_start_end_time in rtp.py at the bottom).
Are you using the start/end record here to get around not having the start_time/end_time working, or is that a feature you would like to be able to do? We also implemented an either/or sort of deal that we used in fan.py plot_fan where the user can put in either a record number or a datetime object in for `scan_index: Union[int, dt.datetime] = 1` then on line 181 in fan.py there is `if isinstance(scan_index, dt.datetime):` to determine what to do (obviously here we would have two keywords to consider). I'm not sure if it might just easier to get the start/end time working properly (but it worked for me anyway so most of the way there) as it makes more sense that a user would know the time they want, but probably not what record in the file that corresponds to.

pydarn/plotting/maps.py
```diff
@@ -1030,3 +1030,117 @@ def plot_potential_contours(cls, fit_coefficient: list, lat_min: list,
...
+        IMF_BY = 'imf_by'
+        IMF_BZ = 'imf_bz'
+        KP = 'kp'
+        HMB_MAX = 'hmb_max'     
+        
+    @classmethod
+    def plot_time_series(cls, dmap_data: List[dict],
+                     parameter: Enum = TimeSeriesParams.NUM_VECTORS,
+                     start_record: int = 0,end_record: int = 1, start_time: dt.datetime = None,
+                     end_time: dt.datetime = None):
```
carleyjmartin, line 1063:
```suggestion
                     end_time: dt.datetime = None, **kwargs):
```
Add **kwargs for matplotlib options in plotting. 
Also add **kwargs to each of the plot.plot calls below to get them into the plots. 
This will allow the user to be able to choose plotting options, like using `color='r'` in the method call will find it's way through to the plot call and you can plot with a red line instead. 
If you want any help with this let me know, this is sometimes a hard one to get right.

carleyjmartin (changes requested):
Hey this is awesome! Thank you so much for taking this on. I don't usually test so quick but you caught me between projects so I had some time today! Most of these comments are just things to keep us all consistent with other plotting functions and recommendations, and as always are just suggestions so if you have better ideas, go for it. If you want any more info/clarification or help with any of the changes just let me know!

I managed to get it working and plotting really nicely!

Other things are: Can we do HMB_MIN as well as MAX? Will this and HMB_MAX work for southern hemisphere data the same way, or will it be flipped because of -ve lat? Might need an abs or if statement here for southern hemisphere data. 

I can also already envision someone wanting IMF_BX and asking why we don't do that, so we can head that off and just add it in too?

Actually, now I think about it, there's quite a few extra scalar parameters that someone might want to plot, in the rtp.py version of plot_time_series, we allowed the user to use any of the parameter names from the file and then when plotting we just pass `parameter` to plot, which might be easier than lots of elifs, then the user can add titles and labels if they want to.  Here are all the params in a map file: https://radar-software-toolkit-rst.readthedocs.io/en/latest/references/general/map/

pydarn/plotting/maps.py
PrestonXPitzer: I'll get working on these changes and hopefully have an updated version out soon 👍 .

PrestonXPitzer:
I think I touched on everything you said @carleyjmartin 
I ended up writing a map specific version of find record that uses a recursive search, which should avoid the issue of not getting any hits, at the cost of not drawing an error when passed datetimes are out of the data range, could implement a manual check fairly easily if it becomes a problem.
I generalized the parameter selection by modifying the params dict to have the correct keys, but this means that the user may have to add their own units unless there is some associated unit with each key:value.
I also moved this dict over to plotting with the other one, and added it to __init__
It now contains all of the values that are useful for analysis (according to Dr. Baker)

carleyjmartin:
This is great, works like a charm.
Tested using this code:
```python
import pydarn
import datetime as dt

file_path = "/Users/carley/Documents/data/maps/20190421.s.map2"
data = pydarn.SuperDARNRead().read_dmap(file_path)
start_time = dt.datetime(2019,4,21,11,0)
end_time = dt.datetime(2019,4,21,13,0)
pydarn.Maps.plot_time_series(data, parameter=pydarn.TimeSeriesParams.CPP,
                             start_time=start_time, end_time=end_time, color='r')
plt.show()
```
and got:
Screenshot 2023-06-28 at 4 19 28 PM


The only thing I would suggest is to run flake8 on the file. There's a few PEP8 styling things we need to tweak up and then we can merge it in!
You can run flake8 by `pip3 install flake8` `flake8 pydarn/plotting/maps.py` and it will print out all the things you might need to tweak to make it PEP8 standard. If something is a bit confusing you can find the guidelines here: https://peps.python.org/pep-0008/
If it finds something on a line you haven't written, don't worry about fixing it, I'll get it before the next release. 
Let me know if you want any help with the styling!

carleyjmartin (approved):

hiyadutuje:
Hi All,

Using the linux computer, after running this code:
import bz2 
import pydarn
import datetime as dt

#file_path = "example_data/20150310.north.map2.bz2"
file_path = "/home/alicreance/pydarn/test/data/maps_n_grids.zip"
data = pydarn.SuperDARNRead().read_dmap(file_path)
pydarn.Maps.plot_time_series(data, parameter=pydarn.Maps.TimeSeriesParams.NUM_VECTORS,
                             start_record=0, end_record=100)
plt.show()
---------------------------------
I am getting the following error:
Error: /home/alicreance/pydarn/test/data/maps_n_grids.zip contains an number of scalars + arrays 1269650028 > block size 524308 at record = 0.
Error: /home/alicreance/pydarn/test/data/maps_n_grids.zip contains an number of scalars + arrays 1269650028 > block size 524308 at record = 0.
..... Will try to read DMap file with read_dmap
 IF THIS FAILS please make an issue on the pyDARNio GitHub repository: https://github.com/SuperDARN/pyDARNio/issues

  File ~/anaconda3/lib/python3.9/site-packages/spyder_kernels/py3compat.py:356 in compat_exec
    exec(code, globals, locals)

  File ~/pydarn/test/336.py:16
    pydarn.Maps.plot_time_series(data, parameter=pydarn.Maps.TimeSeriesParams.NUM_VECTORS,

AttributeError: type object 'Maps' has no attribute 'TimeSeriesParams'

Thank you!

hiyadutuje:
When I change to this:

import bz2 
import pydarn
import datetime as dt

#file_path = "example_data/20150310.north.map2.bz2"
file_path = "/home/alicreance/pydarn/test/data/maps_n_grids.zip"
data = pydarn.SuperDARNRead().read_dmap(file_path)
start_time = dt.datetime(2019,4,21,11,0)
end_time = dt.datetime(2019,4,21,13,0)
pydarn.Maps.plot_time_series(data, parameter=pydarn.TimeSeriesParams.CPP,
                             start_time=start_time, end_time=end_time, color='r')
plt.show()
------------------
I get the following error:

Error: /home/alicreance/pydarn/test/data/maps_n_grids.zip contains an number of scalars + arrays 1269650028 > block size 524308 at record = 0.
Error: /home/alicreance/pydarn/test/data/maps_n_grids.zip contains an number of scalars + arrays 1269650028 > block size 524308 at record = 0.
..... Will try to read DMap file with read_dmap
 IF THIS FAILS please make an issue on the pyDARNio GitHub repository: https://github.com/SuperDARN/pyDARNio/issues

  File ~/anaconda3/lib/python3.9/site-packages/spyder_kernels/py3compat.py:356 in compat_exec
    exec(code, globals, locals)

  File ~/pydarn/test/336.py:18
    pydarn.Maps.plot_time_series(data, parameter=pydarn.TimeSeriesParams.CPP,

  File ~/anaconda3/lib/python3.9/site-packages/pydarn/plotting/maps.py:1144 in plot_time_series
    start_record = cls.find_map_record(dmap_data, start_time)

  File ~/anaconda3/lib/python3.9/site-packages/pydarn/plotting/maps.py:1107 in find_map_record
    start_time, 0, len(dmap_data) - 1)

TypeError: object of type 'method' has no len()
Thanks!

hiyadutuje:
Thank you!
I will try again.
Alicreance
    On Thursday, 16 November 2023 at 17:31:55 GMT+2, Carley ***@***.***> wrote:  
 
  Carley ***@***.***) is not on your Guest List | Approve sender | Approve domain



Thanks for testing, @hiyadutuje
So that first one I think is an error in that you might not be on the right branch to test the new features - it's a little more complicated to test when it's someone else's base repo like this example - but you can try to install the release branch (release/4.0) and test this feature there instead.
Both contain a warning because we don't read in zip files for the read functions but read_dmap will attempt to unzip it, I would unzip the files first either manually or using another python library before passing them to be read.

Issue #312, opened 2023-04-21
Discussion: Ax and CCRS Handling


carleyjmartin:
# Discussion 

**Topic:** Passing axes and projections to plotting functions properly

## Description

At the meeting today we discussed the ability to pass your axes and/or Cartopy projection to the plotting. 
This feature does exist already, however it does not work very well or at all in some cases if you pass a custom axes. 
I think this may take a bit of time properly setting up but would be an excellent update/fix for pyDARN to have in the near future.
If anyone has any example code they have used for the AX, and CCRS keywords or anything they would like to work but doesn't feel free to comment anything below for discussion!

Thanks!

PR #339, opened 2023-06-23
Fixed usage of axes object in fan.py

RemingtonRohel:
* Previously `ax` was allowed as an argument but never used, so current axis was always plotted on.
* Now, user can call plot_fan() and plot_fov() and get the expected output if they specify an axis - i.e., it will actually plot on the axis they passed in.

# Scope 

Behavioural fix in `plot_fan()` and `plot_fov()`

**issue:** N/A

## Approval

**Number of approvals:** *1*

## Test

**matplotlib version**: 3.6.2 and 3.7.1

plot_fan() calls plot_fov(), so call these functions with a fitacf file of your choice and you should see the difference.
```python3
fig, axes = plt.subplots(1, 2, figsize=(12, 6), constrained_layout=True, subplot_kw={'projection': 'polar'})
pydarn.Fan.plot_fan(fit_data, ax=axes[0], scan_index=90, title=True)
pydarn.Fan.plot_fan(fit_data, ax=axes[1],
                    scan_index=90, groundscatter=True, grid=True, zmin=-600, zmax=700,
                    title=True)
plt.show()
plt.close()
```

This is what I get when I plot on the `develop` branch:
develop_fan

And this is what I get on this branch:
fixed_fan


RemingtonRohel:
I should be clear, in the plots I attached I was not plotting the same data in each axis, it was different data but I was only concerned with the grid, title, and FOVs.

pydarn/plotting/fan.py
```diff
@@ -390,11 +390,8 @@ def plot_fov(cls, stid: str, date: dt.datetime,
         -----------
             stid: int
                 Radar station ID
-            ax: matplotlib.pyplot axis
-                Pre-defined axis object to pass in, must currently be
-                polar projection
-                Default: Generates a polar projection for the user
-                with MLT/latitude labels
```
carleyjmartin, line 397: I'm not really sure why you removed this?

carleyjmartin (changes requested):
Disclaimer: Doing this live in a testing tutorial, so will also go more into why this isn't working later. :)

When I ran this the first time, I got an installation error with the hdw files not being included. 

Also got some issues with plotting northern data, see image below:

Screenshot 2023-06-27 at 8 28 21 AM

This is the code I used to plot this:

```python
import pydarn

file = "/Users/carley/Documents/data/20180101.0000.01.rkn.fitacf"

fig, axes = plt.subplots(1, 2, figsize=(12, 6), constrained_layout=True, subplot_kw={'projection': 'polar'})
pydarn.Fan.plot_fan(fitacf_data, scan_index=90)
pydarn.Fan.plot_fan(fitacf_data, ax=axes[1],
                    scan_index=90, groundscatter=True, grid=True, zmin=-600, zmax=700,
                    title=True)
plt.show()
```

pydarn/plotting/fan.py
carleyjmartin: Again, showing people how to test, you can ignore this. :)

carleyjmartin:
Yeah, so I think it's a hemisphere issue. I can plot some southern data, but I'm still missing the grid lines. :/
Screenshot 2023-06-29 at 9 41 35 AM


I can plot northern hemisphere data on a single ax where I don't pass it an ax value, which I think triggers the `if ax is None:` section in `projections.py: axis_polar` method which does have a hemisphere if statement in it, but the contents of it fix southern data being messed up not northern data so I have no idea why northern data is the issue here.

Consistently missing the grid lines on all plots too for some reason, but you're not?


Code testing this time:
```python
import pydarn

file = "/Users/carley/Documents/data/20180101.0000.01.rkn.fitacf"
#file = '/Users/carley/Documents/data/20220108.0801.00.dcn.fitacf'
fit_data = SDarn_read.read_fitacf()

fig, axes = plt.subplots(1, 2, figsize=(12, 6), constrained_layout=True,
                         subplot_kw={'projection': 'polar'})
pydarn.Fan.plot_fan(fit_data, ax=axes[0], scan_index=90, groundscatter=True,
                    title=True)
pydarn.Fan.plot_fan(fit_data, ax=axes[1] scan_index=90, groundscatter=True, 
                    grid=True, zmin=-600, zmax=700, title=True)
plt.show()
```


Honestly, I think we need a whole rewrite of the ax handling for projs. At the moment, we make axes using projs inside the plotting functions, and you can choose geo or polar, BUT if you make an axis beforehand and pass it in, it's not triggering all the code as there's a bunch in if ax is None: section, which was designed to make overplotting on the same axes (once one of these is called first), but only on these two types of axes. This is an issue made a while ago #312

RemingtonRohel:
@carleyjmartin I added a bit of logic to fix "bad" axes objects passed in by users. Basically it just checks whether the ylims are still the default (0.0, 1.0), and if so, then sets them according to the hemisphere. This is what I get plotting RKN data now, using the code from the original PR post: 
fan

carleyjmartin:
Sorted the conflicts with the radar label branch that just got merged into develop - will test properly this afternoon and merge :)

pydarn/plotting/fan.py
```diff
@@ -341,8 +341,8 @@ def plot_fan(cls, dmap_data: List[dict], ax=None, ranges: List = [],
         if ccrs is None:
             azm = np.linspace(0, 2 * np.pi, 100)
             r, th = np.meshgrid(rs, azm)
-            plt.plot(azm, r, color='k', ls='none')
-            plt.grid()
+            ax.plot(azm, r, color='k', ls='none')
+            ax.grid()
```
carleyjmartin, line 345:
```suggestion
```

I'm still getting the grid lines not being shown on fan plots and can only get them to show up by removing this line. Are they appearing on your fan plots?

FOV, Map and grids are all fine for me though, I think I removed this line in another branch too as it was stopping the grid lines... no idea why though.

carleyjmartin:
```suggestion
            ax.grid(True)
```

PR #359, opened 2023-12-15
Major refactor of acf.py

RemingtonRohel:
# Scope 

* Changed all classmethods to staticmethods
* Created staticmethod plot_acf_param() to handle plotting of values, which is called within plot_acfs() as needed to plot real, imaginary, power, and phase of ACF
* Return all axes from plot_acfs() as list
* Plot real and imaginary ACF values on input ax to plot_acfs() if pwr_and_phs False, otherwise use plt.gca()
* Removed some unnecessary internal variables in plot_acfs().
* Proper handling of markers for blanked lags, taking into consideration if marker is a filled marker or not

**issue:** N/A

## Approval

**Number of approvals:** *1*

## Test

**matplotlib version**: 3.7.2

Try running the tutorial from https://pydarn.readthedocs.io/en/main/user/acf/, and using as many of the optional arguments as you can.

### DIFF
--- a/pydarn/plotting/acf.py
+++ b/pydarn/plotting/acf.py
@@ -239,7 +239,7 @@ class ACF:
 
         # Calculate pwr and phs regardless of choice as they are returned
         pwr = np.sqrt(np.square(re) + np.square(im))
-        phs = np.arctan2(im, re)
+        phs = np.rad2deg(np.arctan2(im, re))
         masked_pwr = np.ma.array(pwr, mask=mask)
         masked_phs = np.ma.array(phs, mask=mask)
 
--- a/pydarn/plotting/acf.py
+++ b/pydarn/plotting/acf.py
@@ -7,6 +7,8 @@
 # 2022-05-03: CJM - Added options to plot power and phase of acf
 #                 - change defaults to fit needs
 # 2023-06-28: CJM - Refactored return values
+# 2023-12-15: RAR - Made helper function for plotting ACF values
+#                 - Refactored plot_acfs() to use helper function for plotting
 #
 # Disclaimer:
 # pyDARN is under the LGPL v3 license found in the root directory LICENSE.md
@@ -19,6 +21,8 @@
 
 
 import copy
+
+import matplotlib.markers
 import matplotlib.pyplot as plt
 import numpy as np
 import warnings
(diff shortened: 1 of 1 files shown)


RemingtonRohel:
Made another commit as I realized I had inadvertently changed the phase plotting to plot in radians. It is now back to plotting in degrees.

pydarn/plotting/acf.py
```diff
@@ -253,104 +250,89 @@ def plot_acfs(cls, dmap_data: List[dict], beam_num: int = 0,
...
+        ax.set_ylabel(parameter)
+        ax.set_xlabel('Lag Number')
+        real_dict['data']['line'].set_label('Real')
+        imag_dict['data']['line'].set_label('Imag')
+        if plot_blank:
+            real_dict['data']['blank_line'].set_label('Real Blanked')
+            imag_dict['data']['blank_line'].set_label('Imag Blanked')
+            lim_val = 1.1 * max(np.nanmax(np.abs(re)), np.nanmax(np.abs(im)))
+        else:
+            lim_val = 1.1 * max(np.nanmax(np.abs(masked_re), np.nanmax(np.abs(masked_im))))
```
carleyjmartin, line 283: This line throws a: `TypeError: 'numpy.float64' object cannot be interpreted as an integer`
when I use plot_blank=False

:(

```
rtn = pydarn.ACF.plot_acfs(data, beam_num=5, gate_num=12, normalized=True,
                start_time=datetime(2018, 5, 20, 9, 0), pwr_and_phs=True,
                plot_blank=False)
```

carleyjmartin: Also sorry I've tried to figure out why it's doing this but I dunno

carleyjmartin (commented):
Other than the tiny error, looks great for everything else.

pydarn/plotting/acf.py
carleyjmartin:
```suggestion
            lim_val = 1.1 * max(np.nanmax(np.abs(masked_re),
                                            np.nanmax(np.abs(masked_im))))
```

And the line is too long style wise 👀

RemingtonRohel: Found and fixed the bug! Trying to uses nanmax on a masked array is not a good idea, by default masked_array.max() ignores masked values and all the NaN's were already masked.

PR #363, opened 2024-01-10
Allow user to specify background transparency in RTP calls.

RemingtonRohel:
# Scope 

Allow manual specification of background transparency in RTP plots. This enables overplotting, so that the same axis can have multiple RTP plots on it.

**issue:** N/A

## Approval

**Number of approvals:** 1

## Test

**matplotlib version**: 3.7.2

Try to plot call RTP.plot_range_time() with different data on the same axis object. On develop, it will hide the data from the first call. On this branch, you should be able to see the data behind.

### DIFF
--- a/pydarn/plotting/rtp.py
+++ b/pydarn/plotting/rtp.py
@@ -74,7 +74,8 @@ class RTP():
     @classmethod
     def plot_range_time(cls, dmap_data: List[dict], parameter: str = 'v',
                         beam_num: int = 0, channel: int = 'all', ax=None,
-                        background: str = 'w', groundscatter: bool = False,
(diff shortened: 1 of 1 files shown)


carleyjmartin (changes requested):
So this works great with range time plots alone, but it either shouldn't be passed through to summary plots or needs some more work with the multiple axes and current set up with summary plots.

I'm fine with not passing this option through to summary plots as it's mainly something that is niche to specific types of plots and the summary plots is just that, a summary and if you want something fancier make your own with the range time plots.

RemingtonRohel:
Ah yes I agree. Since the user can't pass in axes anyways to a summary plotting call, that seems reasonable to me!

carleyjmartin (approved):

PR #364, opened 2024-01-11
Lots of little changes to customize Fan plots.

RemingtonRohel:
# Scope 

* Changed projections.py so that a user can pass in an already-defined axis with a projection
* Changed scan.py to find records in a scan by either scan index or datetime
* Refactored fan.py to incorporate the above changes, and allow user to pass in axis object to plot colorbar in.
* Refactored all methods in fan.py to staticmethods

**issue:** N/A

## Approval

**Number of approvals:** 1

## Test

**matplotlib version**: 3.7.2

Try making some fan plots. Test out the `cax` option, `tolerance` option when specifying `scan_index` by a datetime object, passing in an `ax` that is already a cartopy projection, and anything else you think is prudent.

### DIFF
--- a/pydarn/plotting/fan.py
+++ b/pydarn/plotting/fan.py
@@ -49,9 +49,10 @@ from typing import List, Union
 # Third party libraries
 import aacgmv2
 
-from pydarn import (PyDARNColormaps, build_scan, partial_record_warning,
+from pydarn import (PyDARNColormaps, partial_record_warning,
                     time2datetime, plot_exceptions, SuperDARNRadars,
-                    calculate_azimuth, Projs, Coords, Hemisphere)
+                    calculate_azimuth, Projs, Coords,
+                    find_records_by_datetime, find_records_by_scan)
 
 
 class Fan:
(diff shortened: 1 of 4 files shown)


pydarn/plotting/fan.py
```diff
@@ -455,21 +441,24 @@ def plot_fan(cls, dmap_data: List[dict], ax=None, ranges=None,
...
-                cb = ax.figure.colorbar(mappable, ax=ax, extend='max',
-                                        ticks=ticks)
+                extend = 'max'
             else:
-                cb = ax.figure.colorbar(mappable, ax=ax, extend='both',
-                                        ticks=ticks)
+                extend = 'both'
+
+            if cax is None:
+                cax = ax.inset_axes([1.04, 0.0, 0.05, 1.0])
```
carleyjmartin, line 449: This overlaps the '06' label on the plot :(

RemingtonRohel: Oh that's an easy fix, just have to bump the `1.04` value to something larger.

RemingtonRohel:
Huh that's weird, I basically just moved around the code that was already there. What is the value of `scan_indices` in `find_records_by_scan()` when you run the test?

carleyjmartin:
scan_indices is a long list of zeros, matches then returns an empty list

test code:
```python
import pydarn
import datetime as dt

files = ['/Users/carley/Documents/data/20221220.2200.00.rkn.a.fitacf']
for file in files:

    pydarn.Fan.plot_fan(fitacf_data,
                              scan_index=dt.datetime(2022,12,20,22,2),
                              tolerance=dt.timedelta(seconds=120),
                              coastline=True, lowlat=70)

    plt.show()

    pydarn.Fan.plot_fan(fitacf_data, scan_index=1, coastline=True, lowlat=70)
    plt.show()
```

First one plots, second gets the above error for different scan index numbers and different data files with different cpids (and probs scan flag use).

PR #374, opened 2024-02-13
Fix colourbar preference not being preserved

billetd:
# Scope 

This is a very small PR which fixes the behaviour of colourbars when making a convection map. Previously, setting `colorbar=False` when using `plot_mapdata()` would not work if `contour_fill=True`. This is because `plot_potential_contours()` uses a different keyword to control the colourbar for contours, and the one used in `plot_mapdata()` only controls the velocity colourbar.

This PR adds a second keyword to control the potential contour colourbar, `contour_colorbar`, which is set to False as default.

## Approval

1

## Test

```python
import pydarn

mapfile = '20231004.n.map'
SDarn_read = pydarn.SuperDARNRead(mapfile)

pydarn.Maps.plot_mapdata(map_data, record=0, lowlat=50, coastline=True, contour_fill=True, color_vectors=False, colorbar=False, contour_colorbar=False)
```

Gives:
Screenshot 2024-02-13 at 9 57 47 AM

### DIFF
--- a/pydarn/plotting/maps.py
+++ b/pydarn/plotting/maps.py
@@ -133,7 +133,7 @@ class Maps():
                 Default: True
             contour_colorbar: bool
                 Draw a contour colourbar if True
-                Default: False
+                Default: True
             colorbar_label: str
                 The label that appears next to the colour bar.
                 Requires colorbar to be true
--- a/pydarn/plotting/maps.py
+++ b/pydarn/plotting/maps.py
@@ -70,7 +70,8 @@ class Maps():
                      record: int = 0, start_time: dt.datetime = None,
                      time_delta: float = 1,  alpha: float = 1.0,
                      len_factor: float = 150, color_vectors: bool = True,
-                     cmap: str = None, colorbar: bool = True, contour_colorbar=False,
+                     cmap: str = None, colorbar: bool = True, 
+                     contour_colorbar: bool = True,
                      colorbar_label: str = '', title: str = '',
                      zmin: float = None, zmax: float = None,
                      hmb: bool = True, boundary: bool = False,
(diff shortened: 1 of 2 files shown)


pydarn/plotting/maps.py
```diff
@@ -500,7 +503,8 @@ def plot_mapdata(cls, dmap_data: List[dict], ax=None,
                                         lat_shift=lat_shift,
                                         lon_shift=lon_shift,
                                         fit_order=fit_order,
-                                        hemisphere=hemisphere,
+                                        hemisphere=hemisphere
```
carleyjmartin, line 506:
```suggestion
                                        hemisphere=hemisphere,
```

pydarn/plotting/maps.py
```diff
@@ -70,7 +70,7 @@ def plot_mapdata(cls, dmap_data: List[dict], ax=None,
                      record: int = 0, start_time: dt.datetime = None,
                      time_delta: float = 1,  alpha: float = 1.0,
                      len_factor: float = 150, color_vectors: bool = True,
-                     cmap: str = None, colorbar: bool = True,
+                     cmap: str = None, colorbar: bool = True, contour_colorbar=False,
```
carleyjmartin, line 73:
```suggestion
                     cmap: str = None, colorbar: bool = True, contour_colorbar: bool = False,
```

carleyjmartin: Line too long too, bump to next line and tidy up

pydarn/plotting/maps.py
carleyjmartin: Syntax error because of this

pydarn/plotting/maps.py
carleyjmartin: Also think this should be default true, if contours aren't filled it won't plot anyway.
```suggestion
                     cmap: str = None, colorbar: bool = True, contour_colorbar: bool = True,
```

carleyjmartin (changes requested):

carleyjmartin:
Tested with the comma put in and it works as expected.

PR #377, opened 2024-03-14
Change input types, remove len_factor from calculated_fitted_velocities()

billetd:
# Scope 

Small PR to clean up the docstring in `calculated_fitted_velocities()`, and remove the `len_factor' input which isn't used anywhere in the function (probably a holdover from this code was in `plot_mapdata()'. The datatypes for `mlats`, `mlons`, and `fit_coefficient` were changed to arrays from lists, because the code doesn't actually work if you give it a list.

## Approval

**Number of approvals:** 1

## Test

```python
import pydarn
import numpy as np

mapfile = 'path/to/mapfile'
SDarn_read = pydarn.SuperDARNRead(mapfile)

record = map_data[0]  # first record of the file
mlats = np.array([60,70,80])
mlons = np.array([-120,-120,-120])
fit_coefficient = record['N+2']
fit_order = record['fit.order']
lat_shift = record['lat.shft']
lon_shift = record['lon.shft']
lat_min = record['latmin']
hemisphere = pydarn.Hemisphere(record['hemisphere'])

mlats = np.array
velocity, azm = (pydarn.Maps.calculated_fitted_velocities(mlats, mlons, fit_coefficient, hemisphere, fit_order, lat_min))
```

### DIFF
--- a/pydarn/plotting/maps.py
+++ b/pydarn/plotting/maps.py
@@ -585,8 +585,8 @@ class Maps():
                                      and (m != 0) and l**2 + 2 * m - 1) or 0
 
     @classmethod
-    def calculated_fitted_velocities(cls, mlats: list, mlons: list,
-                                     fit_coefficient: list,
+    def calculated_fitted_velocities(cls, mlats: np.array, mlons: np.array,
+                                     fit_coefficient: np.array,
                                      hemisphere: Enum = Hemisphere.North,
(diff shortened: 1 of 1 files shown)


carleyjmartin:
You're correct that calculated_fitted_velocities doesn't need len_factor. 
But you'll need to remove it from the method call earlier too for the code to not break on line 249.

carleyjmartin (approved):

Issue #376, opened 2024-03-07
BUG: Default lowlat used in contour plotting different to default in proj


carleyjmartin:
# BUG 

This bug appears when lowlat is not set in convection maps. Default lowlat generally is 30 in projs, but inside conculate_potentials the default is 60. Contours will only be calculated down the the 60 instead of the hmb or lowest latitude. Very small fix and does not change data that is plotted so minor for now. Normally I would just fix this as it's so minor, but I don't have time right now and will get to it later if no one else wants a cheeky 1-line PR for authorship on a next minor release. 

## Priority

- [x] Minor - typo, incorrect naming, fails on specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

## Example of the bug 

Any call to pydarn.Maps.plot_mapdata with data below 60.

### Attempts

*What have you tried already to fix it?*
Changed int=60 to int=30 in line 904

*What have you tried to get around it?*
For now you can set a lowlat value and it will work fine, don't use the default.

## Potential Solution(s)
See attempt above

PR #378, opened 2024-03-18
FIX: lowlat default value in maps set to 30

carleyjmartin:
# Scope 

This PR amends the lowlat default in calculate_potentials to 30 (which is the default from projections) from 60.

**issue:** #376 

## Approval

**Number of approvals:** 1, 1 liner

## Test

**matplotlib version**:  3.8.2
```python
import pydarn
map_file='data/maps/20220101.n.map'
map_data = pydarn.SuperDARNRead(map_file).read_map()
pydarn.Maps.plot_mapdata(map_data,record=521, 
            lowlat=50, contour_fill=True, reference_vector=250,
            color_vectors=False, coastline=True)
plt.show()
```
Try different and no lowlat kw values - make sure that the contours fill the selected area you have chosen to plot.

### DIFF
--- a/pydarn/plotting/maps.py
+++ b/pydarn/plotting/maps.py
@@ -903,7 +903,7 @@ class Maps():
     @classmethod
     def calculate_potentials(cls, fit_coefficient: list, lat_min: list,
                              lat_shift: int = 0, lon_shift: int = 0,
-                             fit_order: int = 6, lowlat: int = 60,
+                             fit_order: int = 6, lowlat: int = 30,
                              hemisphere: Enum = Hemisphere.North,
                              **kwargs):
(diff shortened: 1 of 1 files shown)


Doreban:
Ran the provided code using matplotlib version 3.8.3.

I tried different `lowlat` values, from 0 through 80, as well as not setting the `lowlat` parameter, and the output plots looked well formatted and scaled.

Here are two example plots, one using `lowlat=55` and one not setting the `lowlat` parameter (so it should be the new default of 30).

pydarn_map_lowlat_55

pydarn_map_lowlat_default

Looks to be working correctly.

Doreban (approved):

Issue #373, opened 2024-02-05
Matplotlib version restriction not sufficient


aburrell:
# BUG 

pydarn fails on import with matplotlib version 3.4.2 due to the lack of a module called "colormaps".  This is fixed by upgrading to matplotlib version 3.8.2. 

## Priority

- [ ] Minor - typo, incorrect naming, fails on specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation 
- [x] Urgent - failure in usage, large misrepresentation

## Information
**python version**: 3.9
**OS**: OS X Big Sur
**matplotlib version**: 3.4.2 (fails); 3.8.2 (succeeds)

## Example of the bug 

With the older version of matplotlib:
```
import pydarn
```

Yields:
```
File ~/Library/Python/3.9/lib/python/site-packages/pydarn/__init__.py:71
     69 from .plotting.color_maps import PyDARNColormaps
     70 from .plotting.projections import Projs
---> 71 from .plotting.rtp import RTP
     72 from .plotting.fan import Fan
     73 from .plotting.grid import Grid

File ~/Library/Python/3.9/lib/python/site-packages/pydarn/plotting/rtp.py:35
     32 import warnings
     34 from datetime import datetime, timedelta
---> 35 from matplotlib import dates, colors, colormaps, ticker
     36 from typing import List
     38 from pydarn import (RangeEstimation, check_data_type, Coords,
     39                     time2datetime, rtp_exceptions, plot_exceptions,
     40                     SuperDARNCpids, SuperDARNRadars,
     41                     standard_warning_format, PyDARNColormaps)

ImportError: cannot import name 'colormaps' from 'matplotlib' (/opt/local/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/matplotlib/__init__.py)
```

### Attempts

*What have you tried already to fix it?* Updated the matplotlib version

### Data Location

No data file necessary

## Potential Bug Location 

pydarn.plotting.rtp.py line 35.  Probably also other places that contain the matplotlib.colormaps.

## Potential Solution(s)

*Please note this is not required.*
1) You could do a try/except for `colormaps` and `cm as colormaps` to be more version agnostic
2) You can increase the minimum supported matplotlib version in setup.cfg (or move these to the pyproject.toml)
 
## Extra Notes

Recommend adding CI tests for different mpl versions to avoid this issue in the future.

carleyjmartin:
Thanks Angeline! I'll update the requirements.

PR #379, opened 2024-03-18
FIX: Increase matplotlib version requirement

carleyjmartin:
# Scope 

This PR increases the required version of matplotlib from 3.3.4 to 3.5.0 due to the change of the colormap module in matplotlib. (I did the fix for this when 3.6.0 came out with the depreciation warning for the color_maps module but didn't change the rtp module, or maybe that was developed later I'm not sure. Anyway,  I don't think we need to have more complicated imports, we can just up the requirements as the version required is nearly 3 years old anyway)

Note that the cm library is still in use elsewhere for other purposes and only 'colormaps' was depreciated and moved from cm.

**issue:** #373 

## Approval

**Number of approvals:** 1

## Test

**matplotlib version**: 3.5.0, 3.4.2, 3.8.3

Install pydarn as you would normally, you can try to install with a matplotlib version lower than 3.5.0 and it should uninstall and install the higher version for you.
Plot an rtp or

### DIFF
--- a/pydarn/plotting/color_maps.py
+++ b/pydarn/plotting/color_maps.py
@@ -1,11 +1,5 @@
 import matplotlib
-import packaging
-
-if packaging.version.parse(matplotlib.__version__) < \
-        packaging.version.parse('3.7.0'):
-    from matplotlib import cm
-else:
-    from matplotlib import colormaps as cm
+from matplotlib import colormaps as cm
 
 
 class PyDARNColormaps():
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,7 +27,7 @@ include_package_data = True
 install_requires =
     pyyaml
     numpy
-    matplotlib>=3.5.0
+    matplotlib>=3.7.0
     aacgmv2
     pydarnio>=1.1.0
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -27,7 +27,7 @@ include_package_data = True
 install_requires =
     pyyaml
     numpy
-    matplotlib>=3.3.4
+    matplotlib>=3.5.0
     aacgmv2
     pydarnio>=1.1.0
 


Doreban:
Did some testing,

on matplotlib-3.8.3:
No issues with importing and I was able to generate RTP plots.

on matplotlib-3.7.0:
No issues with importing and I was able to generate RTP plots.

However, I tried using this branch with matplotlib-3.5.0 and matplotlib-3.6.0, and in both cases I got the following error on importing pydarn.
```
resetting environment variable IGRF_COEFFS in python script
resetting environment variable AACGM_v2_DAT_PREFIX in python script
non-default coefficient files may be specified by running aacgmv2.wrapper.set_coeff_path before any other functions
  File "/home/dgaleschuk/SuperDARN/fit/quick_script.py", line 4, in <module>
    import pydarn
  File "/home/dgaleschuk/SuperDARN/pydarn_testing_env/lib64/python3.11/site-packages/pydarn/__init__.py", line 69, in <module>
    from .plotting.color_maps import PyDARNColormaps
  File "/home/dgaleschuk/SuperDARN/pydarn_testing_env/lib64/python3.11/site-packages/pydarn/plotting/color_maps.py", line 5, in <module>
    class PyDARNColormaps():
  File "/home/dgaleschuk/SuperDARN/pydarn_testing_env/lib64/python3.11/site-packages/pydarn/plotting/color_maps.py", line 6, in PyDARNColormaps
    PYDARN_INFERNO = cm.get_cmap('inferno')
                     ^^^^^^^^^^^
AttributeError: 'ColormapRegistry' object has no attribute 'get_cmap'
```

I was running the following script:

```
sdarn_read = pydarn.SuperDARNRead(file)
fitacf_data = sdarn_read.read_fitacf()

pydarn.RTP.plot_range_time(fitacf_data, beam_num=fitacf_data[0]['bmnum'], range_estimation=pydarn.RangeEstimation.RANGE_GATE)
plt.title("Radar {:d}, Beam {:d}".format(fitacf_data[0]['stid'], fitacf_data[0]['bmnum']))

plt.show()
```

carleyjmartin:
Thanks Draven, yeah I think we need to go up to 3.7.0, even thought he change notes for this are at 3.5.0 🤔

Doreban:
After the new change, if I have matplotlib 3.7.0 installed and install pydarn, just pydarn is installed and it works correctly.

I then downgraded matplotlib to 3.5.0 and attempted to the run the code and it broke as expected. I then tried to reinstall pydarn, and pydarn updated my version of matplotlib to 3.8.3 and everything worked again.

So I would say this is working now.

Doreban (approved):

PR #386, opened 2024-06-28
Record search tolerance

### LINKED ISSUE #385: Tolerence of one second is too low when passing times to plot_fan
# BUG 

#364 introduced a bug(?) when making fan plots by passing a datetime to the `scan_index` keyword.  As `tolerance` works by searching records, not scans, the default of `dt.timedelta(seconds=1)` causes maybe only one beam to be selected for plotting, or non at all depending on the time you pick. Setting `tolerence = dt.timedelta(seconds=60)` fixes this, but this might mess with the plotting if you use an actual index for `scan_index`.

What should the default be? Should we have a different keyword for selecting a time?

## Priority

RemingtonRohel:
# Scope 

Changing the kwargs for `plot_fan()` to clarify which records to plot. Also refactored embargo code for Fan and RTP classes.

closes #385 

## Approval

**Number of approvals:** 1

## Test

Test using `plot_fan()` several different ways: 

* specify `scan_index`
* specify `scan_time`
* specify `scan_time` with a custom `scan_time_tolerance`
* specify nothing at all

Also, test plotting old data (older than 1 year) with a negative CPID to verify that the embargo watermark is displayed correctly. Plot with Fan and RTP classes.

### DIFF
--- a/pydarn/__init__.py
+++ b/pydarn/__init__.py
@@ -52,6 +52,7 @@ from .utils.plotting import TimeSeriesParams
 from .utils.plotting import check_data_type
 from .utils.plotting import time2datetime
 from .utils.plotting import find_record
+from .utils.plotting import determine_embargo, add_embargo
 from .utils.general_utils import GeneralUtils
 from .utils.superdarn_radars import SuperDARNRadars
 from .utils.superdarn_cpid import SuperDARNCpids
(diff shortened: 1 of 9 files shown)


carleyjmartin:
After separating out the embargo fix, all the install errors are gone and I can install fine. I'm still getting the dtime issue when running as:
```python
 rtn = pydarn.Fan.plot_fan(fitacf_data, scan_time=dt.datetime(2020,1,1,0,5),
                              scan_time_tolerance=dt.timedelta(seconds=10),
                              coastline=True, lowlat=30)
```

I think it's trying to send one of the new keywords through kwargs to the aacgmv2 methods this works as expected when plotting in GEO or new MAG projs.

carleyjmartin (approved):

PR #399, opened 2024-07-25
Summary/RTP pots in the style of the VT Superdarn Site

PrestonXPitzer:
# Scope 

This adds a function that produces RTI plots in the same way as they appear on vt.superdarn.org. 

**issue:**

## Approval

Looking for 2 Approvals

## Test

**matplotlib version**: 
```
import pydarn
import datetime
import bz2


def test_vt_rtp():
    with bz2.open('20170101.0000.02.bks.fitacf3.bz2') as f:
        stream = f.read()
    reader = pydarn.SuperDARNRead(stream, True)
    dmap_data = reader.read_fitacf()
    
    pydarn.RTP.vt_web_rtp(dmap_data, beam_num=11)   
    plt.show()
    
if __name__ == '__main__':
    test_vt_rtp()
```
Should produce something like 
image


carleyjmartin:
Thanks Preston!
I've integrated some of the code into the summary plot method as we probably don't need a whole new method just for this little tweak, and this version is specific for your back-end needs!
You can check out the tweaks to the summary plot method in #400

pydarn/plotting/rtp.py
```diff
@@ -1920,6 +1920,496 @@ def plot_coord_time(cls, dmap_data: List[dict], parameter: str = 'v',
...
+        else:
+            h = 1.5+((num_plots)*1.325) 
+        fig = plt.figure(figsize=(11, h))
+
+        start_pos = [0,0,0, 0.60, 0.75, 0.80, 0.84, 0.87, 0.90, 0.92]        
+        if series:
+            for i in range(num_plots):
+                # time-series plots
+                # position: [left, bottom, width, height]
+                if i < 3:   #I have created an abomination
```
carleyjmartin, line 2156: 🤣 


I have no idea how else you would do this tbf.

PR #400, opened 2024-07-26
EHN: Flexibility on Plotting Summary Plot Variables

carleyjmartin:
# Scope 

Following on from #399 
Integrated code into the summary plot method to allow user to print a subset of the allowed vector parameters (user requested)

## Approval

**Number of approvals:** 1

## Test

**matplotlib version**: 3.9.1

```python
import pydarn
import datetime as dt


# Normal data and data that doesnt contain elv anyway so automatically doesn't plot that
files = ["/Users/carley/Documents/data/20211102.0000.00.rkn.a.fitacf",
        "/Users/carley/Documents/data/20231122.0402.00.fhe.fitacf"]

for file in files:
    
    # Single non list value
    a = pydarn.RTP.plot_summary(fitacf_data, beam_num=5, vector_parameters='v')
    plt.show()
    
    # This example should raise an error
    #a = pydarn.RTP.plot_summary(fitacf_data, beam_num=5, vector_parameters=['elv', 'vel'])
    #plt.show()
    
    # Elevation first
    a = pydarn.RTP.plot_summary(fitacf_data, beam_num=5, vector_parameters=['elv', 'v'])
    plt.show()
    
    # Only v in list
    a = pydarn.RTP.plot_summary(fitacf_data, beam_num=5, vector_parameters=['v'])
    plt.show()
    
    # Random other example
    a = pydarn.RTP.plot_summary(fitacf_data, beam_num=5, vector_parameters=['v', 'elv', 'p_l'])
    plt.show()
    
    # Default
    a = pydarn.RTP.plot_summary(fitacf_data, beam_num=5)
    plt.show()
```

Should all plot a summary plot with the given variables in the order given.
I used RKN data for normal data and FHE data to test here to show what would happen if there isn't elevation data anyway (it will just not pot elevation if you ask for it, how it works now).

### DIFF
--- a/pydarn/plotting/rtp.py
+++ b/pydarn/plotting/rtp.py
@@ -872,7 +872,9 @@ class RTP:
                      range_estimation: object =
                      RangeEstimation.SLANT_RANGE,
                      latlon: str = None, coords: object = Coords.AACGM,
-                      **kwargs):
+                     vector_parameters: list = [('p_l'), ('v'),
+                                                ('w_l'), ('elv')],
+                     **kwargs):
         """
         Plots the summary of several SuperDARN parameters using time-series and
         range-time plots. Please see Notes for further description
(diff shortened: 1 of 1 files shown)


Doreban:
Testing on Windows 10, matplotlib version 3.9.1

Summary plots are generated showing only the parameters requested in `vector_parameters`. My only comment is this now leaves a figure with a fair bit of white space if only plotting a single parameter. Is it possible to adjust the figure size based on how many parameters are in `vector parameters`?

summary_plot_1param
summary_plot_2param
summary_plot_all_param

If elevation isn't present in the data, no elevation plot is created as is standard summary plot behaviour. However, since its now possible to be plotting only elevation (passing only 'elv' to `vector_parameters`, the summary plot now crashes if there is no elevation data with the following traceback:
```
  File "C:\Users\Dragon\PycharmProjects\pydarn_dev\test_pr400_flexible_summary_plots.py", line 21, in <module>
    a = pydarn.RTP.plot_summary(fitacf_data, beam_num=5, vector_parameters=['elv'])
  File "C:\Users\Dragon\PycharmProjects\pydarn_dev\.venv\lib\site-packages\pydarn\plotting\rtp.py", line 1303, in plot_summary
    plt.title(cls.__generate_title(x[0], x[-1], beam_num,
UnboundLocalError: local variable 'x' referenced before assignment
```

If you pass in an invalid vector parameter, you get the following exception as expected:
`Exception: Summary plots allow plotting of the following vector parameters only: v (velocity), p_l (Signal to noise ratio), w_l (spectral width), elv (elevation).`

pydarn/plotting/rtp.py
```diff
@@ -1037,32 +1043,52 @@ def plot_summary(cls, dmap_data: List[dict],
...
+        if not isinstance(vector_parameters, list):
+            vector_parameters = [vector_parameters]
+        test_list = [i for i in vector_parameters if i not in allowed_parameters]
+        if len(test_list) > 0 :
+            raise Exception('Summary plots allow plotting of the following '
+                            'vector parameters only: v (velocity), '
+                            'p_l (Signal to noise ratio), '
+                            'w_l (spectral width), '
+                            'elv (elevation).')
+
```
Doreban, line 1077: Would it be good to add a check here to exit out gracefully if `vector_parameters` ends up being empty after the elevation checks?

carleyjmartin:
Thanks Draven, the solution to the white space issue could be cleaner (left note for future work), but I think it's not crazy for the default number of parameters to work with the default figsize, and the user has control over figsize too if they choose to plot less parameters they can make the plot smaller in height themselves - rather than restricting figsize in the code which might give some people issues.

Doreban:
Second crack of tests:

Now if there is no elevation and that was the only parameter requested we get the following exception:
```
Exception: The summary plot needs a list of vector parameters to plot. The file you have chosen to plot may not have elevation data. Summary plots allow plotting of the following vector parameters: v (velocity), p_l (Signal to noise ratio), w_l (spectral width), elv (elevation).
```

If a list is not passed in as the input to `vector_parameters` we now get the following exception:
```
Exception: The summary plot needs a formatted list of vector parameters to plot. E.G. ["v", "p_l"]. Summary plots allow plotting of the following vector parameters: v (velocity), p_l (Signal to noise ratio), w_l (spectral width), elv (elevation).
```

Additionally, now if not all parameters are plotted there is less extra white space:
summary_reduced_whitespace

Also changing the `figsize` doesn't break anything.

Doreban (approved):

Issue #410, opened 2024-12-18
pydarn failing to install in PyHC Environment


sapols:
# BUG 

Your [setup.cfg](https://github.com/SuperDARN/pydarn/blob/develop/setup.cfg) file says pydarn requires `matplotlib>=3.7.0` but I think that might be incorrect. We have pydarn installed in the [PyHC Environment](https://binder.opensci.2i2c.cloud/v2/gh/heliophysicsPy/science-platforms-coordination/pyhc) which uses matplotlib v3.10.0, and pydarn recently started failing to import with this error: 
`ImportError: cannot import name 'InsetPosition' from 'mpl_toolkits.axes_grid1.inset_locator' ([/opt/conda/envs/pyhc-all/lib/python3.10/site-packages/mpl_toolkits/axes_grid1/inset_locator.py](https://hub.binder.opensci.2i2c.cloud/opt/conda/envs/pyhc-all/lib/python3.10/site-packages/mpl_toolkits/axes_grid1/inset_locator.py))`

## Priority

- [ ] Minor - typo, incorrect naming, fails on specific uncommon situation
- [x] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

## Information
**python version**: 3.10
**OS**: Ubuntu
**matplotlib version**: 3.10.0

## Example of the bug 

`import pydarn`
```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
Cell In[25], line 1
----> 1 import pydarn

File [/opt/conda/envs/pyhc-all/lib/python3.10/site-packages/pydarn/__init__.py:60](https://hub.binder.opensci.2i2c.cloud/opt/conda/envs/pyhc-all/lib/python3.10/site-packages/pydarn/__init__.py#line=59)
     58 from .plotting.acf import ACF
     59 from .plotting.power import Power
---> 60 from .plotting.maps import Maps
     61 from .plotting.iq import IQ

File [/opt/conda/envs/pyhc-all/lib/python3.10/site-packages/pydarn/plotting/maps.py:38](https://hub.binder.opensci.2i2c.cloud/opt/conda/envs/pyhc-all/lib/python3.10/site-packages/pydarn/plotting/maps.py#line=37)
     36 from enum import Enum
     37 from matplotlib import ticker, cm, colors
---> 38 from mpl_toolkits.axes_grid1.inset_locator import InsetPosition
     39 from scipy import special
     40 from typing import List

[...]

PR #411, opened 2024-12-19
Replaced deprecated `InsetPosition` import and usage.

RemingtonRohel:
# Scope 

Fix when installing pydarn with matplotlib v3.10

**issue:** #410 

## Approval

**Number of approvals:** 1

## Test

**matplotlib version**: 3.10

Import pydarn and plot a map file **including** the IMF dial (does so by default).

### DIFF
--- a/pydarn/plotting/maps.py
+++ b/pydarn/plotting/maps.py
@@ -29,13 +29,14 @@ Grid plots, mapped to AACGM coordinates in a polar format
 """
 
 import datetime as dt
+
+import matplotlib.axes
 import matplotlib.pyplot as plt
 import numpy as np
 import warnings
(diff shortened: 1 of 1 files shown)


RemingtonRohel:
Testing using this branch and matplotlib v3.10:
inset_axes

Using develop and matplotlib v3.9.2:
develop

However, I did notice a bug with this branch and matplotlib v3.10 when plotting a southern hemisphere map:
inset_axes_2
All those contour lines around 60 degrees are not present when using matplotlib v3.9.2. Not sure why the difference in behaviour, but that is unrelated to the `InsetAxes` fix and so is suitable for a separate issue/PR.

PR #415, opened 2025-03-20
Updating Univ. of Alaska to Penn State change

ksterne:
# Scope 

In looking into pydarn radar info that we're using on the VT website, I noticed that the institution for some radars hasn't changed from University of Alaska to Penn State University.  This has been changed implemented for RST already (https://github.com/SuperDARN/rst/blob/develop/tables/superdarn/radar.dat) and will be put out on the next release.

**issue:**  No issue created with this.

## Approval

**Number of approvals:** Sorry I'm not familiar with this protocol, I would imagine one?

## Test

Test here is making sure I didn't make a typo or break basic functionality.  I imagine running any code that pulls up the radar info to make sure the institution has changed with this branch.


RemingtonRohel:
Looks good to me! I tested pulling the radar information, the Penn State radars all showed up as such for me.

PR #416, opened 2025-05-13
Fix fetching of hdw files.

RemingtonRohel:
# Scope 

Was failing if hdw/ directory not present - simply added a line to make the directory if not present.

**issue:** #410 

## Test

Pull this branch, delete the hdw/ dir, then try importing pydarn from a Python session and confirm that it works.


RemingtonRohel:
I've tested it on my computer, with no `hdw/` dir, an empty `hdw/` dir, and a non-empty `hdw/` dir. All work as expected, I can import pydarn without trouble and those that need hdw files execute a download as expected.

PR #436, opened 2026-02-23
FIX: restrict scipy to >=1.17 and update Legendre code

### LINKED ISSUE #434: BUG/TODO: Scipy v1.17.0
# BUG 

There was a bug in scipy which meant we couldn't use the newer versions for the map plots. This bug in scipy was fixed in the newest release so we need to test and make sure we can use the new version appropriately. And then removes the version restriction.

carleyjmartin:
# Scope 

This PR updates the pyDARN code to use the new version of the scipy.special Legendre function 'assoc_legendre_p_all' instead of the old deprecated 'lpmn' function that was replaced in version 1.15.0, but was broken for a while and is now fixed in the latest 1.17 version of scipy. 
The outputs for the new legendre functions are very different, so to reduce the changes in the code I have reworked the new output to be as similar to the old one as possible.

I tested the new code with scipy 1.15-1.16.3 but I got the issue we had before, so we would need to restrict up to 1.17 for this to work.

With the new code change, it doesn't make sense to keep the old code and if/else it as we still need to have a <1.15 >=1.17 statement anyway, where <1.15 is now 2.5+ years old anyway. 

**issue:** to close #434 

## Approval

**Number of approvals:** 2 for testing please

## Test

**matplotlib version**: 3.10.8
**python version**: 3.11

```python
import pydarn

date_file = '/home/carleyjm/data/map/20251001.n.map'
map_data,_ = pydarn.read_map(date_file)

plt.figure(figsize=(10,10))
pydarn.Maps.plot_mapdata(map_data, record=3, contour_fill = True, color_vectors=False, coastline=True)
plt.show()
```

Comparison between old code and scipy 1.14.0 and new code and 1.17.0:
OLD:
scipy1-14

NEW:
scipy1-17

image

### DIFF
--- a/pydarn/plotting/maps.py
+++ b/pydarn/plotting/maps.py
@@ -638,7 +638,8 @@ class Maps:
         # i is the index of the list
         # x_i is the element of x at ith index
         for i, x_i in enumerate(x):
-            temp_poly = special.lpmn(fit_order, fit_order, x_i)
+            temp_poly = special.assoc_legendre_p_all(fit_order, fit_order, x_i)
+            temp_poly = ((temp_poly[0].T).tolist()[0:fit_order+1],)
             if i == 0:
                 legendre_poly = np.append([temp_poly[0]], [temp_poly[0]],
                                           axis=0)
@@ -973,7 +974,8 @@ class Maps:
         x = np.cos(alpha*theta)
         # Legendre Polys
         for j, xj in enumerate(x):
-            plm_tmp = special.lpmn(fit_order, fit_order, xj)
+            plm_tmp = special.assoc_legendre_p_all(fit_order, fit_order, xj)
+            plm_tmp = ((plm_tmp[0].T).tolist()[0:fit_order+1],)
             if j == 0:
                 plm_fit = np.append([plm_tmp[0]], [plm_tmp[0]], axis=0)
             else:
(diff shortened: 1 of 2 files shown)


alexchartier:
Looks good to me

RemingtonRohel (approved):

RileySylvestre:
**Test**
matplotlib version 3.10.9
scipy version 1.17.1
python version 3.13.13

````
import pydarn

date_file = '/home/riley/repos/pydarn/test/data/test.north.mp'
map_data,_ = pydarn.read_map(date_file)

plt.figure(figsize=(10,10))
pydarn.Maps.plot_mapdata(map_data, record=0, contour_fill = True, color_vectors=False, coastline=True)
plt.savefig("carley_map_test.png", dpi=300)

````
carley_map_test

**Comment by KieranKuch** [comment]
Test - Same results as Riley!
matplotlib version 3.10.9
scipy version 1.17.1
python version 3.13.13

```
import pydarn

date_file = '/home/kieran/repos/pydarn_repo/pydarn/test/data/test.north.mp'
map_data,_ = pydarn.read_map(date_file)

plt.figure(figsize=(10,10))
pydarn.Maps.plot_mapdata(map_data, record=3, contour_fill = True, color_vectors=False, coastline=True)
plt.savefig("map.png", dpi=300)
```

map

Issue #402, opened 2024-08-02
BUG: get_hdw_files() fails on 'unzip' command for Windows.


carleyjmartin:
# BUG 

get_hdw_files() does not work on windows as the unzip command doesn't exist in the same way.

This was only picked up with someone extensively testing, and we haven't come across users having this issue for now. 
If you do have this issue, you can navigate to the hdw repository (`path/to/your/installation/.../pydarn/utils/hdw/`) inside your installation of pyDARN and clone the git repository instead of using this method (`git clone https://github.com/SuperDARN/hdw.git`).

I will not be fixing this for the latest release (v4.1) as it's a very obscure method, but this issue exists to help anyone if they get this error and remind me next year to fix for the next release. 
 

## Priority

- [x] Minor - typo, incorrect naming, fails on specific uncommon situation
- [ ] Major - misrepresentation of data, failure in a common situation 
- [ ] Urgent - failure in usage, large misrepresentation

PR #438, opened 2026-03-05
Improve portability w/ hdw file download

Wtristen:
# Scope 

Modified pydarn.utils.superdar_radars.py 'get_hdw_files' to improve code portability between OS platforms. 
This does not change hdw files themselves or affect any other part of pydarn. The solution replaces external command-line dependencies curl and unzip in get_hdw_files with native python modules urllib.request and zipfile. 

**issue:**
Resolves issue #402 and anticipated portability bugs with use of curl and string path (replaced with os.path) 

## Approval

**Number of approvals:** *number of approvals before one can merge the code*

## Test

**matplotlib version**: 

*provide a list or code to test*

Code to test was uploaded under test/test_hdw.py. To reproduce the issue, use a platform that does not include unzip or curl or one that does not support '/' paths (like old windows I believe); and ensure your hdw directory is empty. 

  - copyrighted (if applicable)


Wtristen:
Sorry if I've goofed something up! This is the first pull-request I've ever attempted and I'm generally new to version control.

pydarn/utils/superdarn_radars.py
```diff
@@ -26,8 +26,11 @@
 from typing import NamedTuple
 from enum import Enum
 from datetime import datetime
-from subprocess import check_call
+#from subprocess import check_call
 
+#new imports
```
carleyjmartin:
```suggestion
```

pydarn/utils/superdarn_radars.py
```diff
@@ -46,35 +49,42 @@ def get_hdw_files(force: bool = True, version: str = None):
...
-
+    # Path should the path where pydarn utils are located, joined with 'hdw'. 
+    hdw_path = os.path.join(os.path.dirname(pydarn.utils.__file__), "hdw")
     if not os.path.exists(hdw_path):
         os.makedirs(hdw_path)
 
     # TODO: implement when DSWG starts versioning hardware files
     if version is not None:
-        raise Exception("This feature is not implemented yet")
+        raise NotImplementedError("This feature is not implemented, Versioned hardware does not exist yet.") #clarified error
```
carleyjmartin:
```suggestion
        raise NotImplementedError("This feature is not implemented, Versioned hardware does not exist yet.") 
```

carleyjmartin (changes requested):
Installed and imported no issue, installed hdw files on installation no issue. 
Removed hdw files and used get_hdw_files() with no issues also.

For now I would remove the test_hdw file, or you can look into pytest and redo it to look similar to the other test files so it works with pytest. It can also be added as a test within test_utils too if that's easier?

Comments and suggestions are just keeping style consistent and tidying. I use flake8 to check pep8 styling if, which you can install with pip and run like flake8 file.py and it will give you places to check.

Great PR!

Wtristen:
I *believe* I've implemented the changes recommended by @carleyjmartin on notation and I've removed the test_hdw.py file that I had previously added in this branch.

carleyjmartin (approved):
Same tests as before, all good. 

I should probably find a windows PC to test this on...

PR #440, opened 2026-03-24
ENH: Option to return full array of data including NaNs from geo calcs

### LINKED ISSUE #439: DISC: Dealing with NaN values from range estimation methods
# Discussion 

**Topic:** range_estimation and coordinates combined

## Description
The range estimation methods often make inf or nan values due to a divide by zero, which when passed on to the coordinates module these values get removed and we don't have to deal with inf or nans when plotting, as that just makes a mess in the plotting functions.

However, this means that we can't directly compare the returned values from the coordinates module as we don't know which range gates have been removed, so we should either return the NaN values with the others, or we should return an array of valid range gates.

I suggest we might have a keyword to return all values, which then would not be used for plotting as we don't return the NaNs as it breaks the plots. 

I don't think we should return 0's.

carleyjmartin:
# Scope 

This PR adds an option to return a full array of lat/lon data from the GEOGRAPHIC and AACGM functions in coordinates.
Often a user does not want to plot the data, but compare the positions in two range estimates (i.e. GSMR and slant range) 
Using this new key word, the user can return a complete lat lon array including NaN values (as inner range gates for GSMR often end up with NaN values due to /0).

Note that this is NOT designed to be used when plotting, including an array with NaNs for position will result in unexpected behavior. You can still plot normally as the default is to remove the NaN values and this hasn't changed.

**issue:** to close #439 

## Approval

**Number of approvals:** 2 - minor change

## Test

**matplotlib version**: 3.10.8

```python
import pydarn
import datetime as dt

data, _ = pydarn.read_fitacf('/home/carleyjm/data/fitacf/20250101.0600.00.sas.a.fitacf')

GSMR_lats, GSMR_lons = pydarn.Coords.GEOGRAPHIC(stid=pydarn.RadarID(data[0]['stid']),
                           rsep = 45,
                           frang=data[0]['frang'],
                           gates=[0,75],
                           date=dt.datetime(2025,1,1,12,0),
                           range_estimation=pydarn.RangeEstimation.GSMR,
                           include_invalid=False)

GSMR_full_lats, GSMR_full_lons = pydarn.Coords.GEOGRAPHIC(stid=pydarn.RadarID(data[0]['stid']),
                           rsep = 45,
                           frang=data[0]['frang'],
                           gates=[0,75],
                           date=dt.datetime(2025,1,1,12,0),
                           range_estimation=pydarn.RangeEstimation.GSMR,
                           include_invalid=True)

SLANT_lats, SLANT_lons = pydarn.Coords.GEOGRAPHIC(stid=pydarn.RadarID(data[0]['stid']),
                           rsep = 45,
                           frang=data[0]['frang'],
                           gates=[0,75],
                           date=dt.datetime(2025,1,1,12,0),
                           range_estimation=pydarn.RangeEstimation.SLANT_RANGE,
                           include_invalid=False)

print('==== GSMR ====')
print(GSMR_lats.shape)
print(GSMR_lons.shape)

print('==== GSMR All ====')
print(GSMR_full_lats.shape)
print(GSMR_full_lons.shape)

print('=== SLANT RANGE ===')
print(SLANT_lats.shape)
print(SLANT_lons.shape)

plt.scatter(SLANT_lats, SLANT_lons, c='r', marker='x')
plt.scatter(GSMR_lats, GSMR_lons, c='b', marker='+')
plt.show()
```
Output:
```
==== GSMR ====
(68, 17)
(68, 17)
==== GSMR All ====
(76, 17)
(76, 17)
=== SLANT RANGE ===
(76, 17)
(76, 17)
```
Figure_1

If the user would like to use range_estimation to calculate positions, this already returns unmodified array of data, including NaNs, plotting removes these in RTP.py but the function can be called to give range estimate values (not lat and lon).

E.G.
```python
range_estimation = pydarn.RangeEstimation.GSMR

rxrise = pydarn.SuperDARNRadars.radars[pydarn.RadarID(data[0]['stid'])].hardware_info.rx_rise_time
y = range_estimation(frang = int(data[0]['frang']),
                     rxrise = rxrise,
                     rsep = int(data[0]['rsep']),
                     nrang=75)
print(y)
```

### DIFF
--- a/pydarn/utils/coordinates.py
+++ b/pydarn/utils/coordinates.py
@@ -32,7 +32,9 @@ from pydarn import (geocentric_coordinates, SuperDARNRadars, RangeEstimation,
 
 
 def geo_coordinates(stid: RadarID, beams: int = None,
-                    gates: tuple = None, **kwargs):
+                    gates: tuple = None,
+                    include_invalid: bool = False,
+                    **kwargs):
     """
     geographic_coordinates calculates the geographic coordinate for a given
     set of gates and beams
@@ -58,12 +60,17 @@ def geo_coordinates(stid: RadarID, beams: int = None,
                                                 **kwargs)
             beam_corners_lats[gate-gates[0], beam] = lat
             beam_corners_lons[gate-gates[0], beam] = lon
-    y0inx = np.min(np.where(np.isfinite(beam_corners_lats[:,0]))[0])
-    return beam_corners_lats[y0inx:], beam_corners_lons[y0inx:]
+    if include_invalid:
+        return beam_corners_lats, beam_corners_lons
+    else:
+        y0inx = np.min(np.where(np.isfinite(beam_corners_lats[:,0]))[0])
+        return beam_corners_lats[y0inx:], beam_corners_lons[y0inx:]
 
 
 def aacgm_coordinates(stid: pydarn.RadarID, beams: int = None, gates: tuple = None,
-                      date: dt.datetime = dt.datetime.now, **kwargs):
+                      date: dt.datetime = dt.datetime.now,
+                      include_invalid: bool = False, 
+                      **kwargs):
     if gates is None:
         gates = [0, SuperDARNRadars.radars[stid].range_gate_45]
     if beams is None:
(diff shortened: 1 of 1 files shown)


Prof-CW:
Hi Carley,
Thanks for adding the information about NaN in the ground scatter lat,lons.
The NaNs come from the calculation of acos() of a number > 1 rather than a divide by 0, i.e.

    ground_scatter_mapped_ranges = (hop/0.5) * Re * np.arccos(-num/den)
    return ground_scatter_mapped_ranges
.

KieranKuch:
Tested on branch ehn/valid-gates-return
with matplotlib version: 3.10.9
and Python version: 3.13.13

Results using the test.fitacf test file:
```
==== GSMR ====
(68, 17)
(68, 17)
==== GSMR All ====
(76, 17)
(76, 17)
=== SLANT RANGE ===
(76, 17)
(76, 17)
```
Verified then that:
`include_invalid=False` removes invalid GSMR gates
`include_invalid=True` preserves the full coordinate arrays, including the NaN values
`SLANT_RANGE` returns the full arrays as expected
Direct range estimation output contains the NaN values for the inner gates

Plotting also still works with filtered data.
test_440_plot

